PR #455, opened 2026-06-17
Fix ledger atomicity and cron cache invalidation

mike840609:
## Summary
- Wrap manual cash transaction creation and manual account balance edits in Prisma transactions.
- Apply holding transaction edits/deletes with guarded row-state checks plus atomic holding quantity deltas.
- Revalidate net-worth before snapshot creation when expired options are swept.
- Add focused route/cron validator coverage and update the enhancement audit docs.

## Verification
- pnpm format:check
- pnpm lint
- pnpm typecheck
- pnpm test:unit
- pnpm build

Note: pnpm build still emits the pre-existing Sentry disableLogger deprecation warning tracked in the audit docs.

### DIFF
--- a/src/app/api/accounts/[id]/route.ts
+++ b/src/app/api/accounts/[id]/route.ts
@@ -32,25 +32,29 @@ export const PATCH = withAuth<IdCtx>(async (request, { params }, userId) => {
   const existingAccount = await prisma.account.findUnique({ where: { id, userId } });
   if (!existingAccount) return failure("Not found", 404);
 
-  // If cashBalance is being updated to a new value, log it as an EDIT transaction
-  if (parsed.data.cashBalance !== undefined) {
-    const diff = new Decimal(parsed.data.cashBalance).minus(existingAccount.cashBalance);
-    if (!diff.isZero()) {
-      await prisma.cashTransaction.create({
-        data: {
-          accountId: id,
-          type: "EDIT",
-          amount: diff,
-          note: body.note || `Manual balance update (${diff.isNegative() ? "" : "+"}${diff})`,
-        },
-      });
+  const { note, ...accountData } = parsed.data;
+  const account = await prisma.$transaction(async (tx) => {
+    // If cashBalance is being updated to a new value, log it as an EDIT transaction.
+    if (accountData.cashBalance !== undefined) {
+      const diff = new Decimal(accountData.cashBalance).minus(existingAccount.cashBalance);
+      if (!diff.isZero()) {
+        await tx.cashTransaction.create({
+          data: {
+            accountId: id,
+            type: "EDIT",
+            amount: diff,
+            note: note || `Manual balance update (${diff.isNegative() ? "" : "+"}${diff})`,
+          },
+        });
+      }
     }
-  }
 
-  const account = await prisma.account.update({
-    where: { id, userId },
-    data: parsed.data,
+    return tx.account.update({
+      where: { id, userId },
+      data: accountData,
+    });
   });
+
   invalidateUserCaches(userId);
   return ok(account);
 });
--- a/src/app/api/accounts/[id]/route.ts
+++ b/src/app/api/accounts/[id]/route.ts
@@ -13,6 +13,8 @@ function invalidateUserCaches(userId: string) {
   revalidateTag(`history:${userId}`, { expire: 0 });
 }
 
+class StaleAccountError extends Error {}
+
 export const GET = withAuth<IdCtx>(async (_request, { params }, userId) => {
   const { id } = await params;
   const account = await prisma.account.findUnique({
@@ -33,27 +35,51 @@ export const PATCH = withAuth<IdCtx>(async (request, { params }, userId) => {
   if (!existingAccount) return failure("Not found", 404);
 
   const { note, ...accountData } = parsed.data;
-  const account = await prisma.$transaction(async (tx) => {
-    // If cashBalance is being updated to a new value, log it as an EDIT transaction.
-    if (accountData.cashBalance !== undefined) {
-      const diff = new Decimal(accountData.cashBalance).minus(existingAccount.cashBalance);
-      if (!diff.isZero()) {
-        await tx.cashTransaction.create({
-          data: {
-            accountId: id,
-            type: "EDIT",
-            amount: diff,
-            note: note || `Manual balance update (${diff.isNegative() ? "" : "+"}${diff})`,
-          },
+  // A manual balance edit logs the difference as an EDIT cash transaction. The
+  // diff must be measured against the balance we actually write over, so the
+  // write is guarded on that prior balance: if a concurrent cash mutation moved
+  // it between our read and the commit, we reject with 409 rather than letting
+  // the EDIT row desync from the stored cashBalance.
+  const nextBalance = accountData.cashBalance;
+  const balanceChanging =
+    nextBalance !== undefined && !new Decimal(nextBalance).equals(existingAccount.cashBalance);
+
+  let account;
+  try {
+    account = await prisma.$transaction(async (tx) => {
+      if (!balanceChanging) {
+        return tx.account.update({
+          where: { id, userId },
+          data: accountData,
         });
       }
-    }
 
-    return tx.account.update({
-      where: { id, userId },
-      data: accountData,
+      const diff = new Decimal(nextBalance!).minus(existingAccount.cashBalance);
+      await tx.cashTransaction.create({
+        data: {
+          accountId: id,
+          type: "EDIT",
+          amount: diff,
+          note: note || `Manual balance update (${diff.isNegative() ? "" : "+"}${diff})`,
+        },
+      });
+
+      const result = await tx.account.updateMany({
+        where: { id, userId, cashBalance: existingAccount.cashBalance },
+        data: accountData,
+      });
+      if (result.count !== 1) {
+        throw new StaleAccountError();
+      }
+
+      return tx.account.findUniqueOrThrow({ where: { id } });
     });
-  });
+  } catch (error) {
+    if (error instanceof StaleAccountError) {
+      return failure("Account changed while updating; please retry", 409);
+    }
+    throw error;
+  }
 
   invalidateUserCaches(userId);
   return ok(account);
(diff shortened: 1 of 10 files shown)


src/app/api/accounts/[id]/transactions/[transactionId]/route.ts
```diff
@@ -19,6 +19,34 @@ function invalidateAccountCaches(userId: string) {
...
+      where: { id: holdingId },
+      data: { quantity: { increment: delta } },
+    });
+    return;
+  }
+
+  if (delta < 0) {
+    const decrement = Math.abs(delta);
+    const result = await tx.holding.updateMany({
+      where: { id: holdingId, quantity: { gte: decrement } },
```
Comment (mike840609) [inline src/app/api/accounts/[id]/transactions/[transactionId]/route.ts:51]: **Float `number` delta against a `Decimal` quantity column (blocking).** `delta` here is pure JS-number arithmetic — it flows from `calculateHoldingQuantityDelta`, which operates on `Number(holdingTx.quantity)` (`src/lib/services/balance.ts:49-56`). Two problems:

1. **Spurious 400.** Closing a fractional position whose delta computes to e.g. `-(0.1 + 0.2)` = `-0.30000000000000004` makes `quantity >= 0.30000000000000004` false against a stored `0.3`, so `updateMany.count === 0` → `NegativeHoldingQuantityError` → `400 "Holding quantity cannot be negative"` on a legitimate close-to-zero.
2. **Ledger drift.** `{ increment / decrement: <float> }` writes a non-exact value into a `Decimal(18,8)` column; over repeated edits the holding quantity diverges from the sum of its transaction ledger — the exact integrity property this PR exists to guarantee.

This also breaks the repo rule in `CLAUDE.md`: *"Use `Decimal` in Prisma operations — never `number` for monetary/quantity values."* Suggest computing the delta as a `Decimal` and passing a `Decimal` to `increment`/`decrement` (and to the `gte` guard).

---

src/app/api/accounts/[id]/transactions/[transactionId]/route.ts
```diff
@@ -62,29 +90,42 @@ export const PATCH = withAuth<TxCtx>(async (request, { params }, userId) => {
...
-          where: { id: holdingTx.holding.id },
-          data: { quantity: nextHoldingQty },
+    let updatedTx;
+    try {
+      updatedTx = await prisma.$transaction(async (tx) => {
+        const result = await tx.holdingTransaction.updateMany({
+          where: {
+            id: transactionId,
+            type: holdingTx.type,
+            quantity: holdingTx.quantity,
```
Comment (mike840609) [inline src/app/api/accounts/[id]/transactions/[transactionId]/route.ts:111]: **Optimistic lock only keys on `type` + `quantity`.** A concurrent edit that changes *only* `note` or *only* `createdAt` leaves `type`/`quantity` untouched, so both racing `updateMany` calls match (`count === 1`) and the second silently clobbers the first — the advertised `409 "Transaction changed while updating"` never fires for those fields. For financial fields this guard is fine, but if the intent is general lost-update protection, a monotonic `updatedAt`/version token in the where-clause would catch any concurrent edit regardless of which column changed.

---

src/app/api/accounts/[id]/transactions/[transactionId]/route.ts
```diff
@@ -62,29 +90,42 @@ export const PATCH = withAuth<TxCtx>(async (request, { params }, userId) => {
...
-        where: { id: transactionId },
-        data: {
-          ...(data.quantity !== undefined && { quantity: nextQuantity }),
-          ...(data.type !== undefined && { type: data.type }),
-          ...(data.note !== undefined && { note: data.note }),
-          ...(data.createdAt !== undefined && { createdAt: new Date(data.createdAt) }),
-        },
+        await applyHoldingQuantityDelta(tx, holdingTx.holding.id, holdingDelta);
+
+        return tx.holdingTransaction.findUniqueOrThrow({
```
Comment (mike840609) [inline src/app/api/accounts/[id]/transactions/[transactionId]/route.ts:126]: **Redundant round-trip (efficiency).** This `findUniqueOrThrow` re-reads a row that was just written by the `updateMany` above, inside the same transaction, purely to obtain a return value. With the optimistic guard you could instead issue a guarded `update({ where: { id: transactionId }, data })` and check for `P2025`/throw — that returns the updated row directly and saves a query on every holding edit (the path is currently 4 DB round-trips: `updateMany` + delta `update`/`updateMany` + `findUniqueOrThrow`).

---

src/app/api/accounts/[id]/transactions/[transactionId]/route.ts
```diff
@@ -19,6 +19,34 @@ function invalidateAccountCaches(userId: string) {
   revalidateTag(`history:${userId}`, { expire: 0 });
 }
 
+class NegativeHoldingQuantityError extends Error {}
```
Comment (mike840609) [inline src/app/api/accounts/[id]/transactions/[transactionId]/route.ts:23]: **Duplicated guard scaffold (maintainability).** The PATCH (lines ~96–128) and DELETE (lines ~217–240) holding branches are two near-identical copies of the same sequence: guarded `updateMany`/`deleteMany` → `count !== 1` throw `StaleHoldingTransactionError` → `applyHoldingQuantityDelta` → an identical `try/catch` mapping `NegativeHoldingQuantityError → 400` and `StaleHoldingTransactionError → 409`. A future change to the stale-row handling or the status mapping has to be made in both places. Worth extracting a single `runGuardedHoldingMutation(...)` helper (co-located with `balance.ts` where the delta math already lives).

---

src/app/api/accounts/[id]/route.ts
```diff
@@ -32,25 +32,29 @@ export const PATCH = withAuth<IdCtx>(async (request, { params }, userId) => {
...
-          type: "EDIT",
-          amount: diff,
-          note: body.note || `Manual balance update (${diff.isNegative() ? "" : "+"}${diff})`,
-        },
-      });
+  const { note, ...accountData } = parsed.data;
+  const account = await prisma.$transaction(async (tx) => {
+    // If cashBalance is being updated to a new value, log it as an EDIT transaction.
+    if (accountData.cashBalance !== undefined) {
+      const diff = new Decimal(accountData.cashBalance).minus(existingAccount.cashBalance);
```
Comment (mike840609) [inline src/app/api/accounts/[id]/route.ts]: **Balance-edit diff is computed from a read taken *outside* the transaction.** `existingAccount.cashBalance` is read at line 32, before this `$transaction`, and the update writes an **absolute** `cashBalance` with no optimistic lock. If a concurrent cash-transaction POST moves the balance between the read and the commit, two things go wrong: the absolute write clobbers that concurrent change (lost update), and the logged `EDIT` row's `amount = newValue − staleBalance` is wrong (e.g. balance 100 → concurrently 150; PATCH sets 200 and logs +100 instead of the true +50). The cash ledger then no longer sums to the stored balance — the same class of bug the holding path was just hardened against. Consider re-reading `cashBalance` inside the transaction (or a conditional `updateMany` guarded on the prior balance) so the EDIT amount and the write are consistent.

---

src/app/api/accounts/[id]/route.ts
```diff
@@ -32,25 +32,29 @@ export const PATCH = withAuth<IdCtx>(async (request, { params }, userId) => {
...
-      await prisma.cashTransaction.create({
-        data: {
-          accountId: id,
-          type: "EDIT",
-          amount: diff,
-          note: body.note || `Manual balance update (${diff.isNegative() ? "" : "+"}${diff})`,
-        },
-      });
+  const { note, ...accountData } = parsed.data;
+  const account = await prisma.$transaction(async (tx) => {
```
Comment (mike840609) [inline src/app/api/accounts/[id]/route.ts]: **Minor: transaction wraps a single statement on the common path.** When `cashBalance` isn't part of the PATCH (renaming an account, changing category, toggling pin/active) this opens an interactive `$transaction` to run a lone `account.update` — begin/commit overhead with nothing to make atomic. The wrapper is only needed on the `cashBalance`-changed branch; the no-EDIT case could fall through to a plain `update`.

---

src/lib/validators.ts
```diff
@@ -24,6 +24,7 @@ export const updateAccountSchema = createAccountSchema
   .extend({
     isActive: z.boolean(),
     isPinned: z.boolean(),
+    note: z.string().max(500).optional().nullable(),
```
mike840609, line 27: **`note` is a non-Account field on the Account schema, and is silently dropped when `cashBalance` isn't also sent.** `Account` has no `note` column, so the PATCH route has to destructure it out (`const { note, ...accountData }`) to keep it from reaching `account.update`. Two consequences:

1. A PATCH carrying only `{ note: "…" }` (no `cashBalance`) validates, gets stripped, results in an empty `account.update`, and the note vanishes with a `200` — the caller believes it saved.
2. Every consumer of `updateAccountSchema` now has to know it must strip `note` before any account write.

Since `note` is really the `CashTransaction` EDIT-row note, a dedicated schema (or a separate request field) would model it at the right altitude rather than overloading account validation.

---

mike840609 (commented):
Code review — overall a solid, net-positive change. It correctly closes the three atomicity gaps it targets (cash create, balance edit, holding edit/delete now commit ledger + totals as one write), adds real optimistic locking on the holding path, and the cron `structuralChanged` change is strictly an improvement (it only *adds* `net-worth`/`accounts` invalidation for the option-sweep case — no new staleness).

Inline comments below. One blocking item and a few "fix didn't go deep enough" items:

- **Blocking — `src/app/api/accounts/[id]/transactions/[transactionId]/route.ts:41`:** `applyHoldingQuantityDelta` passes a JS `number` delta into Prisma `increment`/`decrement` (and the `gte` guard) against a `Decimal(18,8)` column. Float drift can spuriously 400 a legitimate close-to-zero and slowly desync the holding quantity from its ledger — and it breaks the `CLAUDE.md` "never `number` for monetary/quantity values" rule.
- **Asymmetric hardening:** holdings got optimistic locking, but the two sibling cash paths did not.
  - `accounts/[id]/route.ts:39` (inline) — balance-edit diff computed from a read taken outside the transaction.
  - `accounts/[id]/transactions/[transactionId]/route.ts` **cash PATCH branch (~line 162, unchanged so not inline-able):** same lost-update exposure as the holding path now guards against — two concurrent edits of one cash transaction both read `cashTx.amount` outside their transactions and both apply `cashBalance: { increment: delta }`, leaving the row's amount and the account balance inconsistent. Worth applying the same guarded-`updateMany` pattern here.
- Remaining inline notes are lower-severity: lock keys only on `type`+`quantity`, a redundant `findUniqueOrThrow` round-trip, duplicated PATCH/DELETE scaffold, `note` modeled on the Account schema, and a single-statement `$transaction` on the common account-edit path.

---

PR #463, opened 2026-06-18
[codex] Use exchange-rate composite key

mike840609:
## Summary\n\n- Replace ExchangeRate's synthetic string id with the natural fromCurrency/toCurrency composite primary key.\n- Add the matching database migration.\n- Update bulk refresh upserts, change detection, and exchange-rate warmers to use the currency pair.\n\n## Why\n\nThe synthetic id duplicates the currency pair already enforced as unique. The composite key makes the schema and SQL conflict target match the actual domain identity.\n\n## Stack\n\n- Depends on #462.\n- Final PR in the audit cleanup stack.\n\n## Verification\n\n- Prisma schema validation\n- Migration diff reviewed by independent verifier\n- 12 unit test files, 115 tests\n- pnpm build\n- Final tree matches the previously verified audit branch exactly\n- git diff --check\n- pnpm format:check\n- pnpm lint\n- pnpm typecheck


mike840609:
Superseded by #464 after reordering the stack so the shared preview-database migration lands first.

PR #473, opened 2026-06-21
perf(dashboard): render net-worth headline at its real value on first paint

mike840609:
## What

`useCountUp` started at `0` and animated up, so the dashboard's net-worth figure — its largest above-the-fold text — painted **"$0"** until the JS bundle hydrated and the 600 ms count-up ran. On throttled mobile that pushed the real value's paint late and depressed **Speed Index**.

This initializes the hook at the real value and skips the **intro** animation. The roll-from-previous animation on *later* value changes (e.g. a price refresh) is preserved.

## Result

Authenticated dashboard, mobile Lighthouse (median of 3 runs):

| Metric | Before | After |
|---|---|---|
| Speed Index | 3.5 s | **2.3 s** |
| FCP / TBT / CLS / LCP | unchanged | unchanged |

The net-worth figure now appears at its true value on first paint instead of `$0`→animate (verified via Lighthouse filmstrip).

## Notes from the investigation

This started as a broader "raise the mobile Lighthouse score" effort. Measurement (authenticated local prod build + Lighthouse) showed the dashboard is already fast — **85 under real (devtools) throttling, LCP 2.6 s, fully painted by 1.5 s**. The depressed *lab* score is a **Lantern simulation artifact**: it chains the LCP heading's paint behind the full ~756 KB JS bundle because the authed app hydrates behind a stack of client providers. Server-rendering the heading was tried and reverted (no effect — its ancestors are all client providers). The only lever that moves the *simulated* LCP is reducing the initial JS bundle, which is left as a separate, larger effort.

## Scope

- `src/hooks/use-count-up.ts` only (~19 lines).
- No API/behavior change beyond dropping the first-load count-up flourish.

### DIFF
--- a/docs/superpowers/plans/2026-06-21-mobile-initial-load-performance.md
+++ b/docs/superpowers/plans/2026-06-21-mobile-initial-load-performance.md
@@ -1,0 +1,487 @@
+# Mobile Initial-Load Performance Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make the authenticated dashboard paint a useful mobile shell sooner and shorten its critical path without removing charts or increasing Vercel Fluid CPU usage.
+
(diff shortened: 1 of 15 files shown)


mike840609:
Rebased onto latest `origin/master` (linearized — the prior merge commit was flattened) and re-verified locally:

- `pnpm format:check` ✓
- `pnpm lint` ✓
- `pnpm typecheck` ✓
- `pnpm test:unit` ✓ (116 tests)
- `pnpm build` ✓ (`/` still `◐` partial-prerender)

Scope unchanged: the perf fix is `src/hooks/use-count-up.ts`.

PR #483, opened 2026-06-24
fix(price-cache): release claims for symbols missing from a partial fetch

mike840609:
## Problem

In `refreshPricesForHoldings` (price-service.ts), the `refreshingAt` claim is cleared by the bulk upsert **only for symbols that came back with a price** (the `ON CONFLICT ... SET "refreshingAt" = NULL` only touches rows in the upsert).

Claims are explicitly released in two cases:
- total fetch failure (`entries.length === 0`)
- upsert throws

The missing case is a **partial fetch**: claim 10 symbols, Yahoo returns 9 (one bad/transient ticker). The 1 claimed-but-unfetched symbol keeps `refreshingAt` set and stays locked until the 30s `CLAIM_LOCK_TTL_MS` dead-instance expiry — delaying recovery of a symbol whose next retry would succeed.

Impact: low — self-heals after 30s, no data corruption — but it needlessly stalls a transiently-failing ticker.

## Fix

After the upsert, release the gap between claimed and fetched symbols:

```ts
const fetchedSet = new Set(entries.map(([symbol]) => symbol));
await releaseClaims(claimedSymbols.filter((symbol) => !fetchedSet.has(symbol)));
```

## Test

Adds a unit test for the partial-fetch path (two stale symbols claimed, Yahoo returns a price for one only; asserts the release targets the missing ticker and not the fetched one). Total-failure and all-claimed paths were already covered.

`pnpm test:unit` → 124 passed.

### DIFF
--- a/src/lib/services/price-service.ts
+++ b/src/lib/services/price-service.ts
@@ -475,6 +475,12 @@ async function refreshPricesForHoldings(
     updated = entries.length;
     changed = pendingChanged;
     if (changed > 0) revalidateTag("prices", "max");
+
+    // Partial fetch: symbols we claimed but got no price for (bad/transient
+    // ticker) keep their refreshingAt set by the upsert (it only clears rows it
+    // touched). Release them so a retry isn't blocked for the full 30s TTL.
+    const fetchedSet = new Set(entries.map(([symbol]) => symbol));
+    await releaseClaims(claimedSymbols.filter((symbol) => !fetchedSet.has(symbol)));
   } catch (error) {
     errors.push(`Bulk upsert failed: ${String(error)}`);
     // Release claims so the next request can retry immediately.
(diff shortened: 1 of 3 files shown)


mike840609:
**CI fix (Playwright smoke test):** The failing `mobile-plan.spec.ts:60` "desktop Goals and Watchlist remain separate pages" assertion was **not caused by this PR's service-layer change** — a price-cache fix can't affect the `/stocks` onboarding UI.

**Root cause:** the e2e suite runs `fullyParallel` (`workers: 2`, with the `chromium` + `Mobile Chrome` projects sharing one authenticated test user) against a shared, sometimes-cold preview deployment. The desktop assertion keyed off the `/stocks` page title/subtitle (which render from `stocks.title` / `stocks.subtitle` regardless of watchlist contents) but used Playwright's default 5s assertion timeout, so it intermittently lost the race. Reproduced locally: running the suite in parallel surfaces the same flake class on `stocks.spec.ts` and `analysis-populated.spec.ts` on different runs.

**Fix:** gave the first `/stocks` assertion a generous 20s timeout, mirroring `analysis-populated.spec.ts`. Confined to the e2e test; the #483 service fix is untouched. Verified locally: the target test passes 3/3 under the full parallel suite, and `format:check` + `lint` + `typecheck` all pass.

PR #566, opened 2026-07-08
fix: don't let one user's snapshot failure block revalidation for others (#558)

### LINKED ISSUE #558: [bug] One user's snapshot failure fails the whole cron and skips revalidation
## Summary

The daily cron creates all users' snapshots with `Promise.all` — one user's failure marks the whole run failed and, worse, skips the `snapshots`/`history` cache revalidation for every user whose snapshot **did** commit.

## Location

`src/app/api/cron/snapshot/route.ts:143`

## Problem

`Promise.all(users.map(createSnapshot))` rejects on the first failure (e.g. one user's data triggers an error). The in-flight upserts for other users still commit, but control jumps to the catch block, so `revalidateTag("snapshots")` / per-user `history:${id}` (lines 152-154) never run — users see stale history until the next mutation or cron — and the CronRun row + Sentry check-in report total failure, masking the partial success. There's also no per-user error attribution in the log.

## Suggested fix

Use `Promise.allSettled`, log/record which users failed, always run the tag revalidation for the successes, and mark the CronRun degraded-but-ok when only a subset failed.

**Severity:** P2

---

Found by codebase scan 2026-07-06

mike840609:
## Summary

- The daily snapshot cron (`GET /api/cron/snapshot`) created all users' `NetWorthSnapshot` rows with `Promise.all`, so one user's failure rejected the entire batch. In-flight upserts for *other* users still committed to the DB, but control jumped straight to the outer `catch`, which meant `revalidateTag("snapshots")` and the per-user `history:${id}` tag never ran for **any** user — leaving users whose snapshot succeeded looking at stale history until their next mutation or the next day's cron.
- It also masked partial success: the `CronRun` audit row and Sentry check-in reported total failure even when only one user out of many actually failed, with no per-user attribution in the logs.
- Fixed by switching the per-user snapshot loop to `Promise.allSettled` and adding a small pure helper, `classifyCronRunStatus` (`src/lib/cron-run-status.ts`), that classifies the settled results into `ok` / `degraded` / `failed`:
  - `ok` — everyone succeeded.
  - `degraded` — a subset failed; `CronRun.ok` stays `true` (progress was made) but `CronRun.error` records which user(s) failed and why. Revalidation (`snapshots` tag + per-user `history:${id}` tags) now runs for every user whose snapshot actually committed, regardless of the others.
  - `failed` — every user failed; `CronRun.ok = false`, Sentry check-in reports `error`, and the route returns `500`. This is the only case where the batch itself is treated as a hard cron failure — a non-per-user step throwing (price/rate refresh, recurring materialization, etc.) still hits the existing outer `catch` unchanged.
- Each per-user failure is logged individually via `log.error("cron.snapshot.user_failed", { userId, error })` for attribution.
- The recurring cash/investment "catch-up loop" (F6) and the price/rate refresh step are untouched — this PR only touches the per-user snapshot-creation loop and its downstream revalidation/audit logic.

## Test plan

- [x] `pnpm test:unit` — added `tests/unit/cron-run-status.test.ts` (pure classifier: all-success/degraded/all-failed/empty) and extended `tests/unit/cron-snapshot-route.test.ts` with two new cases: (1) one user fails among three — asserts the other two still get their snapshot + revalidation tags, the response is 200 with `degraded: true`, and `CronRun.ok` stays `true` with an error summary naming the failed user; (2) every user fails — asserts a 500 response and `CronRun.ok = false`.
- [x] `pnpm format:check`
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [ ] `pnpm build` — skipped locally (no `DATABASE_URL` in this sandbox); no schema changes were made so this shouldn't be affected.

Fixes #558

https://claude.ai/code/session_01YQk8cJwVFpVmdVCAd45XXB

### DIFF
--- a/src/app/api/cron/snapshot/route.ts
+++ b/src/app/api/cron/snapshot/route.ts
@@ -10,6 +10,7 @@ import { ok, failure } from "@/lib/api-responses";
 import { CRON_SECRET } from "@/lib/env";
 import { log } from "@/lib/logger";
 import { finishSnapshotCronCheckIn, startSnapshotCronCheckIn } from "@/lib/sentry-cron";
+import { classifyCronRunStatus, type CronUserOutcome } from "@/lib/cron-run-status";
 
 function hasValidCronSecret(authHeader: string | null): boolean {
   const expected = Buffer.from(`Bearer ${CRON_SECRET}`);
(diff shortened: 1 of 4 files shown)


mike840609:
Superseded by #563, which was merged directly to master and already ships an equivalent fix (Promise.allSettled, per-user failure logging, revalidation only for successful users, hard-fail only when all users fail). Closing to avoid a duplicate/conflicting fix for the same issue.

PR #635, opened 2026-07-25
perf(docker): slim images 666 MB → 119 MB, build natively per arch, patch HIGH advisories

mike840609:
## Summary

The migration image was built `FROM deps`, so a container whose only job is `prisma migrate deploy` shipped the entire dev dependency tree — Next.js, SWC, TypeScript, Sentry CLI, Playwright. That one layer was 582 MB compressed. This rebuilds it around the Prisma CLI alone, moves both images to alpine, builds each architecture natively instead of under QEMU, and clears the HIGH advisories that surfaced along the way.

| | before | after |
| --- | --- | --- |
| app image | 112 MB | **81 MB** |
| migrations image | 666 MB | **119 MB** |
| app image HIGH/CRITICAL (Trivy) | 32 | **1** |
| migrations HIGH/CRITICAL | 12 | **0** |
| publish job, app image | 18.2 min (QEMU) | native per-arch |
| rebuild after a source edit | 2 min 36 s | 55 s |

## Changes

**`perf(docker)`**
- `migrate` stage builds `FROM base` and installs only the Prisma CLI, pinned to the version the lockfile resolves. It runs through `node` rather than `pnpm exec`, so the container no longer reaches npm for a corepack download before migrating.
- Base moves to `node:24-alpine`. This drops 22 HIGH/CRITICAL OS-package advisories that Debian carries and the app never uses (`perl-base` alone has 4 CRITICAL + 4 HIGH, plus util-linux, gzip, ncurses).
- npm/yarn/corepack removed from both runtime images — clears the CRITICAL node-tar advisory plus `brace-expansion` and `undici`, and leaves no install tooling for anyone who gets a shell. Size is unchanged by this; deleted base-image files only become whiteouts.
- The migrations container runs as non-root (uid 1001), matching the app. Prisma write-checks its engines directory at startup, so ownership is fixed inside the install layer — no extra image size.
- pnpm store and `.next/cache` mounts, scoped per `TARGETPLATFORM`.
- `public/readme-*` excluded from the build context (6.9 MB → 156 KB); also added to `.vercelignore`, since those assets are rendered by GitHub from the repo and never served by the app.

**`ci`** — each architecture builds on its own native runner (`ubuntu-latest` / `ubuntu-24.04-arm`) and pushes by digest; a merge job assembles the tagged manifest lists. QEMU setup removed.

**`fix(deps)`**
- `next` 16.2.10 → 16.2.11: SSRF via rewrites (CVE-2026-64645) and via Server Actions on custom servers (CVE-2026-64649), middleware/proxy bypass in the Turbopack App Router (CVE-2026-64642), Server Actions DoS (CVE-2026-64641).
- `fast-uri` override 3.1.2 → 3.1.4. The pin added in #597 to remediate an advisory had itself become the vulnerable version (CVE-2026-13676, CVE-2026-16221).

## Verification

Full stack brought up from scratch in an isolated compose project after each change: 20 migrations applied to an empty database, re-run reported no pending migrations, login page served with the self-host password field, `/_next/image` returned optimized AVIF/WebP, app healthcheck reached `healthy`, both containers confirmed running as uid 1001. `format:check` clean, `lint` 0 errors, `typecheck` clean, 416 unit tests pass, `pnpm build` succeeds, `actionlint` clean.

Two things caught by the smoke test rather than by inspection:
- Running migrations as non-root initially failed with `Can't write to .../@prisma/engines` — the Prisma CLI write-checks that directory before running and it was root-owned.
- A `sharp` 0.35.0 override (for the remaining advisory) **silently disabled image optimization** — `/_next/image` served the raw 71 KB PNG at every width instead of a 4.5 KB WebP, with nothing in the logs. Next declares `sharp: ^0.34.5`, so pnpm won't link a forced 0.35.0 into Next's own `node_modules` and `require('sharp')` falls into the unoptimized path. Reverted; it is the one remaining Trivy finding. The affected libvips parsing bugs are reachable only through the optimizer, whose remote sources are restricted to `lh3.googleusercontent.com`.

## Notes for review

- The migrate stage reads the Prisma version out of `pnpm-lock.yaml` with `awk` instead of `--frozen-lockfile`, so the CLI version is pinned exactly but its transitive deps resolve fresh at build time. The `docker compose run --rm migrate` step in `docker.yml` guards this. Happy to swap in a committed mini-lockfile if you'd rather have full pinning there.
- `ubuntu-24.04-arm` is the one piece not verifiable locally; it resolves on the first publish run.


mike840609:
## Review verification — one real defect found and fixed (e005778)

Two independent review passes ran over this PR. All CI checks passed on 3692c9c, but a correctness review caught something CI structurally could not.

### The pnpm store cache mount was keyed only by `TARGETPLATFORM`

pnpm caches the *output* of install scripts, so a store shared with a glibc build handed the alpine `migrate` stage a **glibc** Prisma schema-engine. Confirmed on the image built from this branch before the fix:

```
$ ls .../@prisma/engines/ | grep engine
schema-engine-linux-arm64-openssl-3.0.x          # glibc, on an alpine image

$ ./schema-engine-linux-arm64-openssl-3.0.x --version
Error relocating ...: __res_init: symbol not found
```

**It was silent with network access** — Prisma just downloads a replacement engine at container start, which is exactly why every smoke test in this PR passed. On a no-egress network the difference is stark:

```
# before — image built against a cross-libc store
Error: request to https://binaries.prisma.sh/all_commits/.../schema-engine.gz.sha256
  failed, reason: getaddrinfo EAI_AGAIN binaries.prisma.sh

# after
All migrations have been successfully applied.
```

So the pre-fix image carried a per-start egress dependency on `binaries.prisma.sh` — bad for an air-gapped or firewalled self-host, and a startup failure mode with no obvious cause.

**CI was never affected**: GitHub runners start with empty cache mounts, and `type=gha` cache does not carry `type=cache` mount contents. The exposure was local developer builds and any self-hosted runner. That is also why the green Docker smoke test above is not evidence against this.

### Fixes in e005778

- Cache mounts keyed by base image as well as platform (`CACHE_SCOPE` build arg).
- A build-time assertion that the engine matches the image's libc, so a poisoned store **fails the build** instead of shipping a silently-degraded image. Verified: exits 1 against the old image, 0 against the new one.
- **This is also the true cause of the engines-directory permission error** the non-root migration user hit earlier in this PR. Prisma was not write-checking its engines directory — it wanted to write a *downloaded* engine there. The commit corrects that comment.
- Version parsed from `pnpm-lock.yaml` is now validated against a version shape. `test -n` happily accepted the string `peerDependenciesMeta:`, which the awk yields if prisma stops being a direct devDependency.
- Restored tag rules on the per-arch build's metadata step — it pushes by digest, but metadata-action derives OCI labels from tags, so `org.opencontainers.image.version` was being stamped from default rules.
- `pnpm-workspace.yaml` added to the Docker smoke test path filter, since the migrate stage now depends on its `overrides` and `allowBuilds`.

Re-verified end to end after the fix: correct musl engine, all 20 migrations applied **on a no-egress network**, login page asserts both ways (self-host field present, Google button absent, matching the CI assertion), `/_next/image` returns AVIF, actionlint clean.

### Reviewed and deliberately left alone

- **`merge` depends on all four builds**, so a failing migrations build blocks the app manifest. Intentional — `app:v1.2.3` should not publish without a matching `migrate:v1.2.3`.
- **Unpinned transitive deps in the migrate image** — only `prisma` itself is pinned; its ~94 dependencies resolve fresh. Called out in the PR description; happy to add a committed mini-lockfile if preferred.
- **`sharp`** remains the one Trivy finding, for the reason in the PR description.

Verified clean, with evidence: `public/readme-*` unreferenced by app code, npm/corepack removal safe (base entrypoint only does `exec node "$@"`, healthcheck passes), pnpm `overrides`/`allowBuilds`/`minimumReleaseAge` all honored in the migrate image, musl DNS resolves all four outbound API hosts, ICU/timezone output identical to bookworm, `.next/cache` mount does not leak stale `NEXT_PUBLIC_*` values, matrix expands to exactly 4 jobs, digest artifact names cannot collide, and no branch protection references the old `publish` job name.

Issue #511, opened 2026-07-03
[P2] Analysis area (Category Trend) chart plunges to zero at trailing padded month


mike840609:
**Summary**
Analysis area charts plunge to zero at the trailing padded month.

**Location**
`category-trend-chart.tsx:127`.

**Impact / repro**
Category Trend maps empty months to `Number(d[cat] ?? 0)` with no null/`isEmpty` break, so the stacked area collapses to 0 at e.g. Jul 2026 — visible in the screenshot. The assets-vs-liabilities chart got the null-break fix in #481; this one was converted to a stacked area in the same PR but kept `?? 0`.

**Suggested fix**
Apply the same null/`isEmpty` break used by the assets-vs-liabilities chart instead of `?? 0`.

**Source**
Found in site audit of master @ v0.10.0.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015VdCD1kHqhzwh9Y4dDxkwS

mike840609:
Fixed in #527 (v0.10.1) — Category Trend padded months now emit only `monthKey`, and the `<Area>` uses `connectNulls={false}` so the stacked area breaks at gaps instead of plunging to zero. The auto-close keyword didn't trigger on merge; closing manually.

Verified present on master: `category-trend-chart.tsx` connectNulls + `analysis-view.tsx` monthKey-only padding.

PR #688, opened 2026-08-16
perf(analysis): ship precomputed range series instead of raw breakdown (#644)

### LINKED ISSUE #644: [High] /analysis ships the full unbounded snapshot history (incl. per-account breakdown) to the client
## Summary

`/analysis` serialises a user's **entire** snapshot history — including the per-account `breakdown` of every snapshot ever taken — into the RSC payload and hands it to a client component, which then recomputes every chart in the browser on each range switch. Nothing is windowed by the selected range, so both the payload and the client-side work grow without bound as the account ages.

## Evidence

`src/lib/services/analysis-payload-service.ts:25-32` bundles four unbounded reads:

```ts
const [snapshots, cashFlowData, rawHistory, accountCashFlow, investmentCostBasis] =
  await Promise.all([
    getFullNormalizedHistory(userId, baseCurrency),   // every snapshot, no date bound
    getMonthlyCashFlow(userId, baseCurrency),
    getRawHistoryWithBreakdown(userId, baseCurrency), // every snapshot + breakdown JSON
    getAccountMonthlyCashFlow(userId, baseCurrency),  // every (account, month) pair
    getInvestmentCostBasisSummary(userId, baseCurrency),
  ]);
```

`src/lib/services/history-service.ts:299-353` — `getRawHistoryWithBreakdown` selects the heavy `breakdown` column for **all** of the user's snapshots and expands it into one `accountValues` object per day:

```ts
prisma.netWorthSnapshot.findMany({
  where: { userId },                                   // no date filter
  select: { date: true, breakdown: true, baseCurrency: true, createdAt: true },
  orderBy: { date: "asc" },
}),
```

`src/app/(main)/analysis/page.tsx:40-49` passes all of it straight into a client component, and `src/components/analysis/analysis-view.tsx:1` is `"use client"`:

```tsx
<AnalysisView
  snapshots={snapshots}
  cashFlowData={cashFlowData}
  rawHistory={rawHistory}
  accountCashFlow={accountCashFlow}
  ...
/>
```

So everything crosses the server/client boundary. `analysis-view.tsx:22-31` then imports ten pure aggregation functions from `analysis-service.ts` (`aggregateMonthlyChange`, `computeKpis`, `fillMonthRange`, `buildCashFlowBuckets`, `buildCumulativeGrowth`, `aggregateCategoryHistory`, `computePerformanceAttribution`, `computeInvestmentReturn`, `computeInvestmentReturnSeries`, `computeDrawdownSeries`) and runs them client-side — which also pulls that module into the client bundle.

The contrast with the dashboard is instructive: `dashboard-content.tsx:312-320` explicitly notes that `getFullNormalizedHistory` "never selects the heavy `breakdown` column". The analysis path does.

## Failure scenario

A three-year-old account with 12 accounts:

- ~1 100 daily snapshots.
- `rawHistory.snapshots` = 1 100 entries × 12 `accountId → number` pairs. With cuid keys (~25 chars) that is on the order of **400–500 KB of JSON** before compression, for a view whose default range shows 12 months.
- `snapshots` (`NormalizedSnapshot[]`) adds another 1 100 objects carrying `id`, `date`, `createdAt`, three numbers, `baseCurrency`, `label`, `note`.
- `accountCashFlow` adds one entry per (account, month) with any activity — up to 12 × 36.

On a phone that is a multi-hundred-KB RSC payload to parse plus ten aggregation passes over ~1 100 records, repeated on every range toggle (`YTD / 6M / 1Y / All`) because the memo inputs are the full arrays. `getCachedAnalysisPayload` (`revalidate: 300`, tags `net-worth` / `snapshots` / `exchange-rates` / `prices` / `history:${userId}` / `accounts:${userId}`) means a cache miss also re-runs the full `breakdown` scan server-side — and those tags are invalidated by ordinary writes.

## Suggested fix

The range is already a first-class concept in the UI (`analysis-view.tsx` `ranges`), so push it down instead of shipping everything:

1. **Bound the query.** Give `getRawHistoryWithBreakdown` and `getFullNormalizedHistory` a `from` date derived from the selected range (plus one prior month for the baseline that `aggregateMonthlyChange` needs), and make the range part of the `getCachedAnalysisPayload` cache key. `getFullNormalizedHistory` already supports `options.from/to` — `getRawHistoryWithBreakdown` does not.
2. **Aggregate on the server.** The ten functions in `analysis-service.ts` are pure and already take exactly the arrays they need; call them in the server component and send the *derived* series (a few dozen points each) rather than the raw history. That drops both the payload and the client bundle.
3. If "All" must stay unbounded, at minimum drop `id` / `createdAt` / `label` / `note` from the client-bound `NormalizedSnapshot` (only `date` and the three numbers are charted) and pre-collapse `rawHistory` to month-end snapshots — `aggregateCategoryHistory` and `computeInvestmentReturnSeries` only ever use month boundaries anyway.

## Verification method

Static reading of `analysis-payload-service.ts`, `history-service.ts`, `analysis/page.tsx` and `analysis-view.tsx`. Payload sizes above are estimated from the data shapes, not measured against a real account — worth confirming with a production-sized history before sizing the fix.

mike840609:
## Summary

Fixes #644. `/analysis` was shipping the user's **entire** snapshot history — including the per-day `breakdown` of every snapshot (~400–500 KB for a 3-year, 12-account user) — into the RSC payload, and recomputing all ten chart aggregations client-side on every range toggle (YTD/6M/1Y/2Y/All).

Now all five ranges' derived series are **precomputed on the server** and shipped as `seriesByRange`; the client is a thin O(1) `seriesByRange[range]` lookup. Range switching stays instant (fade + persisted range preserved); only the raw `breakdown` payload, the client-side recompute, and the 10-function client bundle are gone.

## Changes

| File | Change |
|---|---|
| `src/lib/services/analysis-series-service.ts` | **new** (server-only) — `computeAnalysisRangeSeries` / `computeAllRangeSeries` replicate the former client `useMemo` chain exactly, once per range |
| `src/lib/services/analysis-payload-service.ts` | payload is now `{ seriesByRange, investmentCostBasis, snapshots, meta }`; cache key gains `locale` (labels are locale-formatted server-side) |
| `src/components/analysis/analysis-view.tsx` | −178 lines: no 10-function imports, no `useMemo` chains, O(1) range lookup |
| `src/components/analysis/analysis-range.ts` | shared `ANALYSIS_RANGES` / `RangeLabel` / `getMonthsForRange` / `pickDefaultRange` (server + client) |
| `src/app/(main)/analysis/page.tsx` | two-phase await — `locale`/`settings` resolve before the payload fetch (no TDZ race in the old `settingsP.then` callback) |
| tests | new `analysis-series-service.test.ts` (parity), `analysis-payload-contract.test.ts` (source contract), extended `analysis-range.test.ts` |

## Verification

- Unit: 811/811 pass (parity tests lock the server computation to the old client math, incl. #511 category padding, cash-flow month window, KPI YTD baseline, drawdown all-time peak)
- `pnpm typecheck` / `lint` / `format:check` clean; `pnpm build` exit 0
- e2e `analysis-populated` PASS (real Chromium, 18-month fixture); e2e `smoke` 3/3 PASS
- Bundle proof: `rawHistory` + all ten aggregation fn names have zero hits in client chunks (`.next/static/chunks`) — server-only now

## Deliberate deviations from the issue's suggested fix

1. **Precompute all 5 ranges in one cache entry** instead of range-in-cache-key — preserves the instant client-side range switching that's a codebase convention (dashboard trend chart does the same), avoids cache fragmentation.
2. **No `from` param on `getRawHistoryWithBreakdown`** — dead code under precompute-all (YAGNI).
3. **No month-end collapse of `rawHistory`** — verified it would change attribution/return results (`computePerformanceAttribution`/`computeInvestmentReturn` use the first in-range daily snapshot; `computeInvestmentReturnSeries` uses per-month first/last).
4. **`snapshots` still shipped** — required by the mobile `#history` tab (`HistoryView`); ~3× smaller than the removed breakdown payload, same data the `/history` page already ships.

Known accepted staleness: range boundaries / default range freeze at cache-fill (≤300 s revalidate, same class as the existing payload cache) and use server TZ for the first-visit default.

## Commits

- `refactor(analysis): extract shared range constants, label type, and default picker`
- `feat(analysis): aggregate per-range chart series on the server`
- `perf(analysis): ship precomputed range series instead of raw breakdown (#644)`


mike840609:
Follow-ups from this review that were deliberately left out of the fix commit are tracked in #689 (locale cache key, `src/lib` ↔ `src/components` layering, redundant `meta` fields, `rangeLabelKey` duplication).

PR #710, opened 2026-08-23
fix: version client storage and stabilize timers

### LINKED ISSUE #696: [P3] Version client storage keys before persisted-state migrations
## Problem
Several client persistence keys have no schema/version boundary, including:

- `privacy-mode` in `src/components/layout/privacy-mode-context.tsx`
- `asset-tracker:color-schema` in `src/components/layout/color-schema-context.tsx`
- `asset-tracker:sidebar-collapsed` in `src/components/layout/sidebar.tsx`
- the persisted range key in `src/hooks/use-persisted-range.ts`

## Impact
There is no immediate performance impact and the current values work. The risk appears when a value changes shape, semantics, or ownership: old browser state can be silently interpreted as current state, and generic keys can collide with another version or application.

## Proposed outcome
Adopt a consistent versioned key convention for client persistence, or store a version field alongside persisted values. Define the migration behavior for existing users before changing any persisted value format.

## Acceptance criteria
- All app-owned local/session storage keys follow one documented versioning convention.
- Existing values either remain compatible or have an explicit migration/ignore path.
- Invalid and unknown versions fall back safely to defaults.
- Tests cover current values, unknown versions, and migration behavior where applicable.

### LINKED ISSUE #697: [P3] Stabilize client timer lifecycles for cooldown and freshness badges
## Problem
The audit found two small but related client timer inefficiencies:

1. `StockTrackerView` recreates its cooldown interval every second because the effect depends on `now`:
   `src/components/stocks/stock-tracker-view.tsx:737-742`
2. Every `FreshnessBadge` creates its own 30-second interval, and the component also schedules a second immediate timeout when `timestamp` changes:
   `src/components/ui/freshness-badge.tsx:36-49`

## Impact
The current impact is low: these do not represent a user-visible correctness bug. They do create avoidable timer setup/cleanup and periodic state updates, especially on pages with several freshness badges.

## Proposed outcome
Make cooldown timing use a stable interval while a cooldown is active, and centralize or otherwise reduce the freshness clock so multiple badges share timer work. Preserve the current visible timing and SSR-safe first render.

## Acceptance criteria
- The stock cooldown interval is not recreated on every one-second tick.
- Freshness badges do not each need an independent periodic timer when they share the same clock.
- Timestamp changes still update the displayed freshness promptly.
- Timers are always cleaned up on unmount and when no longer needed.
- Add focused fake-timer tests for cooldown expiry and freshness updates.

mike840609:
## Summary

- Fixes #696 by centralizing app-owned Web Storage keys under asset-tracker:v1, validating values, and lazily migrating recognized legacy keys with safe defaults.
- Fixes #697 by keeping cooldown timers stable and sharing one freshness clock across badges with cleanup and prompt timestamp refresh.

## Verification

- pnpm test:unit — 111 files, 963 tests passed
- pnpm typecheck
- pnpm lint
- pnpm format:check
- git diff --check

Closes #696
Closes #697

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -13,6 +13,20 @@ The current application stack is Next.js 16, React 19, Prisma 7, PostgreSQL, Tai
 - `prisma/schema.prisma` — persistent data model
 - `prisma/migrations/` — ordered production schema history
 
+## Client persistence
+
+App-owned `localStorage` and `sessionStorage` keys use
+`asset-tracker:v<version>:<name>`. The current version is `v1`, and
+`src/lib/client-storage.ts` is the registry for every key and its legacy alias.
+
+Readers accept only values recognized by the current consumer. When no current key exists, a
+recognized legacy value is copied to the `v1` key and the legacy key is removed. Invalid current or
+legacy values and keys from unknown versions are ignored so the consumer's normal default applies.
+Storage access and migration failures are non-fatal; a failed copy leaves the legacy value intact.
+
+The sidebar cookie remains `asset-tracker:sidebar-collapsed` because it seeds server rendering and is
+not Web Storage. The `theme` key is owned by `next-themes` and is outside the app-owned convention.
+
 Authentication uses NextAuth.js with JWT sessions. Non-Vercel deployments may use a built-in single-owner password, Google OAuth, or both; Vercel production remains Google-only. `src/auth.config.ts` stays runtime-safe while `src/auth.ts` contains server-only adapter and credentials-provider configuration.
 
 ## Public Demo boundary
--- a/src/components/layout/color-schema-context.tsx
+++ b/src/components/layout/color-schema-context.tsx
@@ -15,6 +15,7 @@ import {
   APP_ICON_FAVICON_RADIUS,
   createAppIconSvg,
 } from "./app-icon";
+import { CLIENT_STORAGE_KEYS, readClientStorage, writeClientStorage } from "@/lib/client-storage";
 
 export type ColorSchema = "emerald" | "anthropic" | "ocean" | "violet" | "amber" | "rose";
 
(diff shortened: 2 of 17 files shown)


mike840609 (commented):
## Review: ready to merge ✅

Merges cleanly against `master` (`cb0441a`) and all 11 checks are green — unit, lint, typecheck, format, build, bundle size, PostgreSQL integration, and both Playwright suites.

The `client-storage.ts` versioning-with-legacy-migration and the shared clock in `timer-lifecycles.ts` are well-built. `startCooldownTicker` also quietly fixes the old `[cooldownUntil, now]` effect that re-created its interval on every tick.

### Non-blocking notes

**1. `pwa-install-hint.tsx` — behavior change for storage-blocked browsers**

The old code suppressed the hint entirely when storage threw:

```js
try { hasBeenShown = window.localStorage.getItem(STORAGE_KEY) === "1"; } catch { return; }
```

`readClientStorage` now swallows the throw and returns `null`, so `hasBeenShown` is `false` and the flow continues; `markShown` then also fails silently. On a browser with storage blocked (Safari private mode — exactly the audience this hint targets) the toast will reappear on **every** page load.

`pwa-install-prompt.tsx` already returned `false` on throw, so only the hint regressed here.

**2. `getPrivacyModeSnapshot` performs writes**

It's a `useSyncExternalStore` getSnapshot, so it is called during render, and `readClientStorage` can `setItem`/`removeItem` as part of the legacy migration. The returned value is stable so React won't misbehave, but a side-effecting getSnapshot is worth avoiding — the one-time migration could run in the provider's effect instead.

**3. `subscribeToFreshnessClock` — latent double-timer**

Timers start whenever `freshnessListeners.size === 1`. Passing a listener reference that is already subscribed would start a second timer pair and leak the first. Not reachable from any current call site, since React creates a fresh closure per effect.

Also worth noting: when a legacy value fails the `allowed` check, `readClientStorage` returns `null` but leaves the legacy key in storage. Harmless, just some cruft that never gets cleaned up.

None of these block the merge.

---

PR #711, opened 2026-08-23
perf: finish mobile PWA audit improvements (#705)

### LINKED ISSUE #705: Mobile PWA performance improvements from codebase audit
## Summary

Codebase audit for mobile PWA performance identified 6 improvement opportunities, ranked by impact.

### 1. Service worker has no caching (`public/sw.js`)
Current SW is network-only — every launch re-downloads all assets and the app is completely unusable offline. The `fetch` listener intercepts every same-origin GET but adds zero value.

**Proposal:**
- Static assets (`_next/static/*`, icons, fonts): **stale-while-revalidate**
- HTML routes: **network-first + cache fallback** (show last screen offline)
- API routes: keep network-only (financial data should not be cached)
- If not implementing caching, remove `fetch` handler entirely to eliminate interception overhead
- Update `tests/unit/service-worker.test.ts` accordingly

### 2. Pull-to-refresh triggers React re-render on every touchmove
`src/components/layout/pull-to-refresh.tsx:45-56` calls `setPull()` on every touchmove event (60–120 Hz on high-refresh devices), cascading re-renders through context consumers.

**Proposal:** Direct DOM transform via ref + `requestAnimationFrame` batching during touchmove; only setState at touchend when refresh is triggered. Also replace fragile `document.querySelector("main")` (line 29) with a ref or context-based reference.

### 3. Goals hub loads all four tabs' data in one blocking `Promise.all`
`src/app/(main)/goals/page.tsx` awaits goals + accounts + projections + stocks + calendar entries + earnings before returning any HTML. On slow mobile connections TTFB equals the slowest request.

**Proposal:** Stream non-default tab data into independent `<Suspense>` boundaries, or fetch inactive-tab data client-side on first tab activation.

### 4. Recharts duplicated across route bundles (~+150KB gzip)
Noted in `src/components/analysis/analysis-view.tsx:108` comment: Turbopack duplicates recharts across route bundles.

**Proposal:** Run `pnpm analyze` to confirm duplication across dashboard/history/analysis; consolidate imports through a shared barrel module with dynamic import if confirmed.

### 5. Sticky header `backdrop-blur-md` continuous GPU cost
`src/components/layout/mobile-header.tsx:47` — blur filter runs every frame during scroll on low-end devices. Consider reducing blur radius or degrading to solid background while scrolling.

### 6. Root layout preconnects Vercel scripts unconditionally
`src/app/layout.tsx:110` preconnects `va.vercel-scripts.com` even when analytics are disabled. Move inside the `enableVercelInsights` conditional.

## Already Good

- Charts lazy-loaded (`lazy-charts.tsx`, `lazy-analysis-charts.tsx`)
- Toaster lazy-imported
- Mobile hub proxy-level UA redirect prevents client flash
- Hydration-safe viewport hook via `useSyncExternalStore`
- Correct `Cache-Control: no-cache` on `/sw.js`

mike840609:
## Summary

- Cache only same-origin static GET assets in the service worker with SWR; documents, RSC, API, and private data remain network-only.
- Move pull-to-refresh drag updates to rAF/ref DOM writes with explicit cleanup and preserved gesture behavior.
- Reduce mobile header blur and gate Vercel preconnect hints.
- Make cache writes best-effort and protect navigation/document boundaries.

Closes #705

## Verification

- Focused unit suite: 4 files, 41 tests passed.
- Mobile Chrome E2E: 5 passed, 5 skipped.
- Typecheck and production build: passed.

### DIFF
--- a/public/sw.js
+++ b/public/sw.js
@@ -38,8 +38,10 @@ function isCacheableRequest(request) {
 async function refreshStaticAsset(request) {
   const response = await fetch(request);
   if (response.ok) {
-    const cache = await caches.open(STATIC_CACHE);
-    await cache.put(request, response.clone());
+    caches
+      .open(STATIC_CACHE)
+      .then((cache) => cache.put(request, response.clone()))
+      .catch(() => {});
   }
   return response;
 }
--- a/public/sw.js
+++ b/public/sw.js
@@ -38,9 +38,10 @@ function isCacheableRequest(request) {
 async function refreshStaticAsset(request) {
   const response = await fetch(request);
   if (response.ok) {
+    const cacheResponse = response.clone();
     caches
       .open(STATIC_CACHE)
-      .then((cache) => cache.put(request, response.clone()))
+      .then((cache) => cache.put(request, cacheResponse))
       .catch(() => {});
   }
   return response;
--- a/public/sw.js
+++ b/public/sw.js
@@ -30,6 +30,7 @@ function isStaticAsset(pathname) {
 
 function isCacheableRequest(request) {
   if (request.method !== "GET") return false;
+  if (request.destination === "document" || request.mode === "navigate") return false;
   const url = new URL(request.url);
   return url.origin === self.location.origin && isStaticAsset(url.pathname);
 }
--- a/public/sw.js
+++ b/public/sw.js
@@ -1,10 +1,60 @@
-// Minimal network-only service worker.
-// Satisfies Chrome's PWA installability requirement (needs a fetch listener)
-// without adding caching — same-origin GET requests pass through to the network.
+const STATIC_CACHE = "astt-static-v1";
+const STATIC_CACHE_PREFIX = "astt-static-";
+
 self.addEventListener("install", () => self.skipWaiting());
-self.addEventListener("activate", (e) => e.waitUntil(self.clients.claim()));
-self.addEventListener("fetch", (e) => {
-  if (e.request.method === "GET" && new URL(e.request.url).origin === self.location.origin) {
-    e.respondWith(fetch(e.request));
+self.addEventListener("activate", (event) =>
+  event.waitUntil(
+    (async () => {
+      const cacheNames = await caches.keys();
+      await Promise.all(
+        cacheNames
+          .filter((name) => name.startsWith(STATIC_CACHE_PREFIX) && name !== STATIC_CACHE)
+          .map((name) => caches.delete(name)),
+      );
+      await self.clients.claim();
+    })(),
+  ),
+);
+
+function isStaticAsset(pathname) {
+  return (
+    pathname.startsWith("/_next/static/") ||
+    pathname.startsWith("/icons/") ||
+    pathname === "/manifest.webmanifest" ||
+    pathname === "/favicon.ico" ||
+    pathname === "/icon" ||
+    pathname === "/icon.svg" ||
+    pathname === "/apple-icon"
+  );
+}
+
+function isCacheableRequest(request) {
+  if (request.method !== "GET") return false;
+  const url = new URL(request.url);
+  return url.origin === self.location.origin && isStaticAsset(url.pathname);
+}
+
+async function refreshStaticAsset(request) {
+  const response = await fetch(request);
+  if (response.ok) {
+    const cache = await caches.open(STATIC_CACHE);
+    await cache.put(request, response.clone());
   }
+  return response;
+}
+
+self.addEventListener("fetch", (event) => {
+  if (!isCacheableRequest(event.request)) return;
+
+  event.respondWith(
+    (async () => {
+      const cache = await caches.open(STATIC_CACHE);
+      const cached = await cache.match(event.request);
+      if (cached) {
+        refreshStaticAsset(event.request).catch(() => {});
+        return cached;
+      }
+      return refreshStaticAsset(event.request);
+    })(),
+  );
 });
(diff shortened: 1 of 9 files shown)


mike840609 (commented):

### Blocking

**1. Service worker: `clone()` after `await` (`public/sw.js`)**

```js
async function refreshStaticAsset(request) {
  const response = await fetch(request);
  if (response.ok) {
    caches.open(STATIC_CACHE)
      .then((cache) => cache.put(request, response.clone()))  // ← too late
      .catch(() => {});
  }
  return response;
}
```

`response.clone()` runs inside the `caches.open(...)` callback, but `response` is returned synchronously to `respondWith`. By the time the callback runs the browser may already be consuming the body, so `clone()` throws `TypeError: Response body is already used` — and the `.catch(() => {})` swallows it, so **the cache silently never gets written**. Whether it works depends on microtask timing, which is the worst failure mode: it will look fine locally and populate nothing in production.

Clone synchronously before returning:

```js
const copy = response.clone();
caches.open(STATIC_CACHE).then((cache) => cache.put(request, copy)).catch(() => {});
return response;
```

**2. Merging this breaks a test that landed on `master` after the branch point**

#706 rewrote `pull-to-refresh.tsx` and added `tests/unit/pull-to-refresh.test.ts`, which imports `dampedPull` and `applyPullTransform`. This PR deletes both exports and does not touch that test file (it doesn't exist on this branch). Post-merge, typecheck and unit tests fail unless the resolution also rewrites that suite.

**3. The pull-to-refresh conflict is a competing rewrite, not a merge**

This branch redesigns the context API from `getMain()`/`getIndicator()` to `mainRef`/`indicatorRef` and reimplements the drag loop — on top of a file `master` already rewrote for the same goal in #706 ("eliminate per-touchmove React re-renders in pull-to-refresh"). This needs a deliberate decision about which implementation wins, not a mechanical conflict resolution.

**4. No CI has run on this branch**

The only checks here are the Vercel preview jobs (triggered by `deployment_status`). `ci.yml` and `e2e.yml` both trigger on `pull_request`, but neither has a workflow run against `codex/issue-705-mobile-pwa-audit`. Lint, typecheck, unit tests, build, bundle size, PG integration, and the local Playwright suite all have **zero** results here.

**5. Preview E2E failure is unexplained**

`calendar.spec.ts:38` (desktop CRUD) timed out at 60s — the gridcell for August 12, 2031 never appeared. This PR doesn't touch calendar code, and three sibling previews running concurrently passed, so shared-preview-DB interference is plausible. But this is the only one of the four PRs introducing a service worker that intercepts asset fetches, so I would not write it off as a flake without a clean isolated re-run.

### Note on the caching policy

`/_next/static/` is content-hashed so SWR is safe there. `/icons/`, `/manifest.webmanifest`, `/favicon.ico`, `/icon`, `/icon.svg`, and `/apple-icon` are not hashed, so the first load after a change serves stale — acceptable for icons, just worth being deliberate about.

The exclusions look right: documents, navigations, RSC, and API all stay network-only, and the `activate` handler cleans up old `astt-static-` caches.

---

mike840609:
## Review follow-up

- Cloned network responses synchronously before opening the cache, so stale-while-revalidate writes cannot race response consumption.
- Rebased onto the latest `master` and reconciled the pull-to-refresh implementation with #706's API/exports and tests.
- Preserved the network-only boundaries for documents, navigations, API/private data, cross-origin requests, and non-GET requests.
- Re-ran the relevant checks: unit 973/973, lint, typecheck, format, build, calendar E2E, and CI/preview smoke checks passed.

mike840609 (commented):
## Re-review at `c3ff32e` — all blockers resolved ✅

Full CI ran this time and is green: unit, lint, typecheck, format, bundle size, PostgreSQL integration, and both Playwright suites.

### Previous findings

**1. Service worker `clone()` after `await` — fixed.** The clone is now taken synchronously before the response is returned:

```js
const cacheResponse = response.clone();
caches.open(STATIC_CACHE).then((cache) => cache.put(request, cacheResponse)).catch(() => {});
return response;
```

**2. Broken `tests/unit/pull-to-refresh.test.ts` — fixed.** `dampedPull` and `applyPullTransform` are back, and the unit suite is green.

**3. Competing PTR rewrite — resolved the right way.** The branch dropped the parallel reimplementation and now builds on master's version. What's left is a genuine, focused improvement: `refreshingRef` removes `refreshing` from the effect dependency array, so the touch listeners are no longer torn down and re-attached every time the refresh state flips. That's a real fix, and much easier to review than the original rewrite.

**4. Preview E2E — passing.** The `calendar.spec.ts` failure did not reproduce on this run, which supports the shared-preview-DB interference theory from the first review rather than anything in this diff.

### One cosmetic leftover

`applyPullTransform(main, indicator, 0, false)` sets the indicator to `translate(-50%, 0px)`, while the new `resetPullStyles` uses `INDICATOR_HIDDEN_Y` (`-44px`). The component no longer calls `applyPullTransform` for the reset path, so the `0px` branch is now reached only during an upward drag (where opacity is 0, so nothing is visible) and by the unit test. No visual impact — just two slightly different notions of "reset" living in the same file. Worth collapsing whenever that test gets touched next.

The rest — gated Vercel preconnect, `backdrop-blur-sm`, the declarative indicator transform — all reads cleanly. The declarative `style` on the indicator and the imperative writes during drag don't fight each other, since React only re-renders on the `refreshing` flip.

Good to merge from my side.

---

PR #712, opened 2026-08-23
perf: lazy-load mobile Plan tab data (#703)

### LINKED ISSUE #703: perf: lazy-load mobile Plan tab data on /goals
## Problem

`/goals` currently starts all Plan-tab data requests in `GoalsContent`, even for mobile users whose initial tab is Watchlist. This makes the first PWA/mobile load wait on goals, projection, calendar, and earnings data that is not rendered yet. Related to #701.

## Proposed direction

- Render the mobile Plan shell and Watchlist with only the data needed for the active tab.
- Load Goals, Projections, and Calendar data when the corresponding tab is selected, or use an equivalent server/client route split.
- Keep desktop `/goals` behavior and deep links intact.
- Avoid fetching calendar earnings from the external provider on the mobile critical path unless Calendar is opened.

## Acceptance criteria

- Mobile `/goals` initial request does not fetch inactive-tab data.
- Switching tabs loads the required data and preserves loading and error states.
- Desktop behavior and URL/hash deep links remain unchanged.
- Mobile TTFB, LCP, and RSC payload are compared before and after.

## References

- #701
- `src/app/(main)/goals/page.tsx`
- `src/components/goals/goals-view.tsx`

mike840609:
## Summary

- Select only the active mobile Plan panel's authenticated server readers.
- Preserve desktop Goals behavior, tab query/hash deep links, and Calendar month/date state.
- Reconcile UA/viewport differences without permanent loading skeletons and keep i18n/service work parallel.
- Add route/data-selection regressions and keep loading state accessible.

Closes #703

## Verification

- Full unit suite: 103 files, 931 tests passed.
- Focused Goals tests and Mobile Chrome Plan/Calendar E2E: passed.
- Typecheck, formatting, and diff checks: passed.

### DIFF
--- a/messages/en-US.json
+++ b/messages/en-US.json
@@ -125,6 +125,7 @@
     "delete": "Delete",
     "cancel": "Cancel",
     "save": "Save",
+    "loading": "Loading...",
     "saving": "Saving...",
     "deleting": "Deleting...",
     "asset": "Asset",
--- a/messages/zh-TW.json
+++ b/messages/zh-TW.json
@@ -125,6 +125,7 @@
     "delete": "刪除",
     "cancel": "取消",
     "save": "儲存",
+    "loading": "載入中...",
     "saving": "儲存中...",
     "deleting": "刪除中...",
     "asset": "資產",
--- a/src/app/(main)/goals/page.tsx
+++ b/src/app/(main)/goals/page.tsx
@@ -7,6 +7,7 @@ import { getProjectionData } from "@/lib/services/projection-service";
 import { getCachedTrackedStocks } from "@/lib/services/stock-watch-service";
 import { getLocale, getMessages, getTranslations } from "next-intl/server";
 import { NextIntlClientProvider } from "next-intl";
+import { headers } from "next/headers";
 import { pickMessages } from "@/lib/i18n-utils";
 import { LargeTitleHeading } from "@/components/layout/large-title-heading";
 import { GoalsView } from "@/components/goals/goals-view";
@@ -24,9 +25,13 @@ import {
   type CalendarEarningsItem,
 } from "@/lib/services/calendar-earnings-data";
 import { rateLimitSubjectCheckWithPrune } from "@/lib/rate-limit";
+import { isMobileUserAgent } from "@/lib/mobile-hub-route";
+import { parseMobilePlanTab, type MobilePlanTab } from "@/components/goals/mobile-plan-tabs";
 import { Card, CardContent } from "@/components/ui/card";
 import { Skeleton } from "@/components/ui/skeleton";
-import type { GoalWithProgress, SerializedAccount } from "@/lib/types";
+import type { GoalWithProgress, SerializedAccount, SerializedCalendarEntry } from "@/lib/types";
+import type { ProjectionData } from "@/lib/services/projection-service";
+import type { SerializedTrackedStock } from "@/lib/services/stock-watch-service";
 
 const CLIENT_NAMESPACES = [
   "goals",
@@ -42,7 +47,7 @@ const CLIENT_NAMESPACES = [
 ];
 
 type GoalsPageProps = {
-  searchParams: Promise<{ month?: string; date?: string }>;
+  searchParams: Promise<{ month?: string; date?: string; tab?: string }>;
 };
 
 function PanelSkeleton() {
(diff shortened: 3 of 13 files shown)


mike840609 (commented):

The implementation is coherent and the tests are updated consistently with the new `?tab=` URLs (`proxy.test.ts`, `mobile-plan.spec.ts`, `calendar.spec.ts`). The blockers here are mostly decisions rather than defects.

### Blocking

**1. This reverts the Suspense streaming that #706 just added**

`GoalsDeferredData` and `PanelSkeleton` are deleted and replaced with server-side UA sniffing plus tab-scoped data selection. That is a second, competing solution to the same problem that landed on `master` a few hours earlier ("perf(goals): stream non-default tab data via Suspense boundaries"). Merging this silently undoes that change — it needs an explicit call on which approach wins, not a conflict resolution.

**2. No CI has run on this branch**

The only checks are the Vercel preview jobs. Neither `ci.yml` nor `e2e.yml` has a run against `codex/issue-703-goals-lazy-data`, so lint, typecheck, unit, build, bundle size, PG integration, and the local Playwright suite have zero results. Given this PR changes route data selection and touches three test suites, that matters.

**3. Merge conflict on `src/app/(main)/goals/page.tsx`** against the #706 version.

### Design trade-offs worth confirming

**UA sniffing makes `/goals` User-Agent dependent.** `headers().get("user-agent")` on the server vs. viewport-based `isMobile` on the client. Any mismatch — narrow desktop window, iPad "request desktop site" — trips the reconcile effect and fires `router.replace`, costing an extra RSC round trip on every load for those users.

**Tab switching is now network-bound.** `handleTabChange` moved from `window.history.replaceState` to `router.replace`, so each mobile tab switch is a server round trip that renders `MobilePlanLoading`, where previously every panel was already hydrated. That is the intended trade for the lazy loading, but switching back to an already-viewed tab also re-fetches.

### Minor

- `parseMobilePlanTab` falls back to `"watchlist"` rather than `undefined`, so `/goals?tab=garbage` on desktop loads watchlist data server-side (an unnecessary `getCachedTrackedStocks` call) and then self-corrects with a second navigation via the reconcile effect.
- Tab state is now duplicated in both `?tab=` and `#hash`. `handleTabChange` also writes `#watchlist`, where the old code deliberately left the bare URL hashless.
- The `goalsDataP!` non-null assertions in `GoalsContent` are safe as written, but they're load-bearing on the `activeTab` narrowing directly above them — easy to break in a later edit.

---

mike840609:
## Review follow-up

- Rebased onto the latest `master` and preserved #706's Suspense streaming and desktop eager/deferred reader behavior.
- Mobile server reads now follow the active tab: Goals/accounts, watchlist, projections, or calendar data only.
- Invalid desktop `tab` values no longer select a mobile fallback; query/hash URLs are canonicalized while legacy deep links remain supported.
- Removed the load-bearing non-null assertion and kept loading/accessibility states covered.
- Verification: focused and full unit tests, lint, typecheck, format, build, PostgreSQL, bundle, and Playwright smoke checks passed; Vercel preview smoke also passed.

mike840609 (commented):
## Re-review at `f8a1a65` — all blockers resolved ✅

Full CI ran this time and is green across the board.

### Previous findings

**1. Reverting #706's Suspense streaming — resolved.** `Suspense` and `PanelSkeleton` are retained and the tab-scoped data selection now runs *inside* the deferred boundary. The two approaches are combined rather than one replacing the other, which was the outcome I was hoping for.

**2. Duplicated `?tab=` + `#hash` state — resolved.** The query is now the single canonical selector, hashes are supported for existing links and stripped after the first client render, and `watchlist` carries no `tab` param at all so the default URL stays clean. `getMobileHubUrl` follows the same rule.

**3. `?tab=garbage` on desktop — resolved.** `requestedTab = isMobile && tab !== undefined ? parseMobilePlanTab(tab) : undefined` means desktop ignores the param entirely, and the `!isMobile` branch of `getGoalsPanelData` still loads the full set (projections, stocks, calendar). Desktop behavior is genuinely preserved now.

**4. `goalsDataP!` non-null assertions — gone.** `getGoalsPanelData` is extracted as a plain function with an explicit `parseDateOnly` null check replacing the old `parseDateOnly(from)!`, and it's directly covered by `goals-page-data-selection.test.ts`. Much easier to follow than the nested ternary chain.

**5. Unnecessary `router.replace` round trips — reduced.** The reconcile effect now uses `history.replaceState` when `loadedTab === desiredTab` and only falls back to `router.replace` when the server actually has to load a different tab.

### Accepted trade-offs

Both of these are inherent to the design and I read the update as a deliberate confirmation, not something to revisit:

- **UA sniffing** makes `/goals` User-Agent dependent, and a UA/viewport mismatch still costs one reconcile navigation.
- **Tab switching is network-bound** — `handleTabChange` goes through `router.replace`, so each mobile tab switch is a server round trip behind `MobilePlanLoading`. That is the point of lazy loading.

Minor nit, no action needed: `getMobileHubClientRedirectUrl` / `getMobileHubRedirectUrl` still take a `hash: \`#${string}\`` parameter that is now only used to derive the tab and never appears in the returned URL. The name outlived its meaning.

Good to merge from my side.

---

PR #713, opened 2026-08-23
perf: remove disabled Sentry client bundle (#702)

### LINKED ISSUE #702: perf: avoid shipping Sentry client bundle when browser telemetry is disabled
## Problem

`src/instrumentation-client.ts` statically imports `@sentry/nextjs`, so Sentry appears in the shared client bundle even when `NEXT_PUBLIC_SENTRY_DSN` is unset. The current build shows a shared Sentry-containing chunk of approximately 436 KB raw and 136 KB gzip. This affects every route and PWA startup.

## Proposed direction

- Gate the client import at build time when no DSN is configured, or split/defer it while preserving early error capture for builds that enable Sentry.
- Keep router instrumentation and error reporting working when Sentry is configured.
- Add a bundle-size regression check for DSN-disabled builds.

## Acceptance criteria

- No-Sentry builds do not ship the Sentry browser SDK in shared route chunks.
- Sentry-enabled builds retain current telemetry behavior.
- Route JavaScript transfer and mobile Web Vitals are verified before and after.

## References

- #701
- `src/instrumentation-client.ts`

mike840609:
## Summary

- Lazy-load browser Sentry only when a public DSN exists, including both client error boundaries.
- Preserve enabled initialization options and router-transition forwarding.
- Add a no-browser-Sentry bundle assertion to CI.

Closes #702

## Verification

- No-DSN build + bundle assertion: passed.
- Enabled DSN build: Sentry chunk emitted as expected.
- Focused Sentry tests: 5/5; typecheck and formatting: passed.


mike840609 (commented):

The bundling goal is sound and the `--assert-no-sentry` CI gate is a good guard against regressing it. The problem is what happens when a DSN **is** configured: the change degrades the reporting path it is meant to preserve.

### Blocking

**1. `global-error.tsx` capture is now fire-and-forget across a dynamic import**

```js
const sentryClient = process.env.NEXT_PUBLIC_SENTRY_DSN
  ? import("../instrumentation-client-sentry")
  : undefined;
...
void sentryClient?.then(({ captureException }) => captureException(error, ...));
```

At this point the root layout has thrown and the document is being replaced. Reporting the error now requires fetching the Sentry chunk, initializing the SDK, then capturing and flushing — all after the crash. Global errors are precisely the case where you least want an async chunk load between the failure and the report. Same pattern in `(main)/error.tsx`, with lower stakes.

Worth considering: keep a synchronous capture path when the DSN is present, and gate only the *no-DSN* build on the dynamic import.

**2. `onRouterTransitionStart` no longer starts the navigation span synchronously**

```js
export const onRouterTransitionStart: typeof captureRouterTransitionStart = (...args) => {
  void sentryClient?.then(({ onRouterTransitionStart }) => onRouterTransitionStart(...args));
};
```

Sentry's SDK expects this hook to run synchronously at the start of a navigation to open the span. Deferring it into a `.then()` means the span opens after the chunk resolves — on the first navigation it will be mistimed or missed entirely.

**3. No CI has run on this branch**

Only the Vercel preview jobs are present. Neither `ci.yml` nor `e2e.yml` has a run against `codex/issue-702-sentry-client-bundle` — which is notable given this PR *adds a new CI step* that has therefore never executed here.

**4. Merge conflict** on `tests/unit/sentry-initialization.test.ts` against `master`.

### Minor

- The guard test only greps for the literal string `import * as Sentry from "@sentry/nextjs"`. A named import (`import { captureException } from "@sentry/nextjs"`) would reintroduce the bundle and still pass. The `--assert-no-sentry` bundle check is the real guard; this test adds little.
- `assertNoSentry` re-reads each file once per marker via `markers.some((m) => readFileSync(file, "utf8").includes(m))` — up to 4 full reads per file. CI-only, so cosmetic, but reading once and testing all four markers is free.

---

mike840609:
## Review follow-up

- Rebased onto the latest `master` and resolved the Sentry test conflict.
- Added a build-time Turbopack alias that selects a synchronous Sentry client module when a DSN is configured and a no-op module for no-DSN builds.
- Error-boundary capture and `onRouterTransitionStart` are synchronous when enabled.
- Strengthened the no-Sentry guard for namespace, named/default, dynamic, and require imports, while reading each bundle file once.
- Verification: focused Sentry tests, no-DSN bundle assertion, DSN build, unit, lint, typecheck, format, integration, bundle, and Playwright checks passed.

mike840609 (commented):
## Re-review at `4ddb02e` — all blockers resolved ✅

Full CI ran this time and is green, including the Docker self-host smoke test (picked up because `next.config.ts` changed).

### Previous findings

Both async-capture blockers are gone, and the fix replaces the mechanism rather than patching it. Instead of a conditional `import()`, the branch now uses a build-time module alias:

```ts
const sentryClientModule = process.env.NEXT_PUBLIC_SENTRY_DSN
  ? "./src/instrumentation-client-sentry.ts"
  : "./src/instrumentation-client-noop.ts";
...
turbopack: { resolveAlias: { "@/instrumentation-client-sentry": sentryClientModule } }
```

That gets both properties at once:

**1. `global-error.tsx` capture is synchronous again.** Plain `import { captureException } from "../instrumentation-client"` and a direct call — no chunk fetch between the crash and the report, which was my main concern given the document is being torn down at that point.

**2. `onRouterTransitionStart` is synchronous again.** A static re-export of `Sentry.captureRouterTransitionStart`, so the navigation span opens at transition start as the SDK expects.

**3. The `--assert-no-sentry` gate does run on PRs.** I got this wrong in the first review — I read the step as landing in the push-only `build` job. It's actually in `bundle-size`, which is gated on `pull_request` and passed green here. So the no-DSN build is empirically verified to contain no Sentry markers, which is also the real proof the alias works.

### Two things to keep in mind

- **The alias is Turbopack-only.** `resolveAlias` lives under `turbopack`, which is fine on Next 16 where Turbopack is the default for `next build`. But a `--webpack` build would resolve `@/instrumentation-client-sentry` to the real module and quietly reintroduce the bundle. The `bundle-size` gate would catch it, so this is a note rather than a risk.
- **The noop's signatures aren't type-checked against the real module.** TypeScript resolves the import through `tsconfig` paths to the real Sentry file and knows nothing about `resolveAlias`, so `instrumentation-client-noop.ts` is only checked standalone. If `captureRouterTransitionStart`'s signature changes upstream, the noop can drift without a typecheck error.

Good to merge from my side.

---

PR #714, opened 2026-08-23
perf: defer Analysis ranges and below-fold charts (#704)

### LINKED ISSUE #704: perf: defer Analysis range data and below-the-fold charts
## Problem

Analysis currently fetches and serializes all range series plus snapshots before rendering, and `AnalysisView` mounts eight dynamic chart components. Dynamic imports split code but do not defer all offscreen work. This makes `/analysis` expensive, especially in an installed PWA.

## Proposed direction

- Render the default range first and fetch or compute additional ranges on demand.
- Lazy-mount below-the-fold charts with viewport or visibility gating and stable skeletons.
- Avoid duplicate Recharts vendor code across dashboard, history, and analysis route chunks.

## Acceptance criteria

- Initial `/analysis` RSC payload and JavaScript transfer are reduced.
- The default chart becomes interactive without waiting for all charts and ranges.
- Switching ranges and scrolling into charts preserves behavior and accessibility.
- Layout shift does not exceed the current skeleton dimensions.
- Mobile LCP and INP, plus route JavaScript size, are measured before and after.

## References

- #701
- `src/app/(main)/analysis/page.tsx`
- `src/components/analysis/analysis-view.tsx`

mike840609:
## Summary

- Ship only the default Analysis range initially while preserving snapshots and cost basis.
- Add an authenticated Demo-allowed range API backed by shared cached inputs.
- Load missing ranges on demand with abort/retry handling.
- Defer Movement/Composition charts with an 800px IntersectionObserver boundary and reattach it across range changes.
- Add the exact Demo route policy entry and retry-state regression.

Closes #704

## Verification

- Full unit suite: 104 files, 917 tests passed.
- Focused Analysis tests: 95 tests passed; typecheck, lint, formatting passed.
- Populated Chromium Analysis E2E: 4 passed, 4 skipped.
- Build and route/RSC artifact measurements recorded in the report.


mike840609 (commented):

### Blocking

**1. Clock-consistency regression in `getCachedAnalysisRangeSeries`**

`computeAnalysisRangeSeries` documents the invariant explicitly:

> `now` freezes at cache-fill time — accepted staleness (same class as the rest of the cached payload). **Pass one shared `now` for the whole fill so every range resolves against the same instant.**

The new call site doesn't pass it:

```js
computeAnalysisRangeSeries(snapshots, rawHistory, cashFlowData, accountCashFlow, range)
//                                                                    ↑ no `now`
```

so it falls back to the `now = new Date()` default. This PR also deletes the comment on the old call site that warned about exactly this failure — *"every range and the default range must agree on 'now', even when a fill straddles midnight."* The default range and an on-demand range can now resolve against different instants, and each instant is then frozen into a 300s cache.

**2. `unstable_cache` key omits the inputs**

```js
unstable_cache(
  async () => computeAnalysisRangeSeries(snapshots, rawHistory, cashFlowData, accountCashFlow, range),
  ["analysis-range-series", userId, baseCurrency, range],
  { revalidate: 300, ... },
)
```

The closure captures `snapshots` / `rawHistory` / `cashFlowData` / `accountCashFlow`, but none of them are in the cache key. `analysis-inputs` and `analysis-range-series` have independent 300s TTLs that start at different moments, so there is a window where an on-demand range is computed from older inputs than the default range rendered beside it. The shared tags cover tag-driven revalidation, but not the TTL skew.

**3. The `analysis-view.tsx` conflict is not mechanical**

#708 moved locale resolution to `useLocale()` inside the chart components and removed the prop plumbing. This branch still threads `locale` from `page.tsx` → `AnalysisView` → `KpiTiles` / `LazyAssetsLiabilitiesChart` / `LazyCategoryTrendChart`. Taking this side of the conflict produces a typecheck failure — the `locale` entry in `Props`, all three `locale={locale}` props, and the `page.tsx` plumbing all have to be dropped during resolution.

**4. No CI has run on this branch**

Only the Vercel preview jobs are present; neither `ci.yml` nor `e2e.yml` has a run against `codex/issue-704-analysis-deferred-work`. For a change that adds a new authenticated API route and reshapes a cached service, that's the gap that matters most here.

### Minor

- In the `!series` loading branch the IntersectionObserver sentinel sits near the top of the page, so `sentinel.getBoundingClientRect().top <= window.innerHeight + 800` is true immediately and `showDeferredCharts` flips to `true` while data is still in flight. The 800px below-fold deferral therefore never applies to on-demand ranges — every chart mounts at once when the data lands. Not harmful, but the deferral only actually does work on the initial default-range render.
- `seriesByRange` widening to `Partial<Record<RangeLabel, ...>>` is the right call, and the `drawdownSeries` guard and abort/retry handling both look correct — `series` identity stays stable per range across `setSeriesCache` updates, so the memo and the fetch effect don't churn.

---

mike840609:
## Review follow-up

- Merged the latest `master` and reconciled the #708 `useLocale()` migration by removing obsolete locale prop plumbing.
- Added one shared `cacheFillAt` timestamp for default and on-demand ranges, and included it in range cache keys so derived ranges cannot reuse a different input fill.
- Deferred the IntersectionObserver setup until the selected range data exists, preserving the 800px boundary for on-demand ranges.
- Verification: 971 unit tests, 109 focused Analysis tests, lint, typecheck, format, build, Analysis E2E, integration, bundle, and Playwright checks passed.

mike840609 (commented):
## Re-review at `17b7fca` — all blockers resolved ✅

Full CI ran this time and is green.

### Previous findings

**1 & 2 — the clock and the cache key are fixed by one mechanism.** `cacheFillAt` is stamped *inside* the cached inputs function, so it's the fill timestamp and stays stable for the life of that cache entry:

```ts
const cacheFillAt = new Date().toISOString();   // inside unstable_cache
```

That single value then does both jobs:

- `getCachedAnalysisPayload` and `getCachedAnalysisRangeSeries` both derive `now` from `new Date(cacheFillAt)`, so the default range and every on-demand range resolve against the same instant. The midnight-straddle invariant from the `computeAnalysisRangeSeries` doc comment holds again, and the comment is back on the call site.
- `cacheFillAt` is part of the range-series key — `["analysis-range-series", userId, baseCurrency, cacheFillAt, range]` — so a new inputs fill can't have its derived range entries served from the previous fill. That closes the TTL-skew window cleanly, and it's a better fix than hashing the inputs themselves.

**3. The `locale` conflict — resolved toward master.** No `locale` references remain in the `analysis-view.tsx` diff; the branch adopted #708's `useLocale()` approach and dropped the prop plumbing. Typecheck confirms.

**4. Sentinel tripping during load — fixed.** The deferred-charts effect now guards on `!series`:

```ts
if (showDeferredCharts || !hasData || !series) return;
```

so it no longer measures a sentinel sitting near the top of the loading skeleton. The 800px below-fold deferral now actually applies once data lands, instead of being pre-tripped.

Nothing further from me. Good to merge.

---

PR #715, opened 2026-08-24
feat(login): turn /login into the public landing surface

mike840609:
## Why

`astt.app` has no public page. Every anonymous visit is redirected to `/login` (`PUBLIC_ROUTES`, `src/proxy.ts:17`), which shows only a sign-in card — so directory listings (awesome-selfhosted, AlternativeTo, OpenAlternative) and HN/GitHub traffic land on a password box with no pitch.

Rather than add a marketing site or a second route (a landing at `/` would fight `(main)/page.tsx` and the PPR root), this puts the pitch on the page visitors already reach.

## What

- `/login` renders a two-column grid on `lg+`: pitch left, sign-in card right. Mobile keeps the card first, pitch below.
- Pitch = headline, subline, three value points, the Docker quick-start from the README, and the product screenshot (`lg`-only, lazy, so mobile pays nothing).
- The Suspense fallback now mirrors the two-column shell, so the streamed content doesn't jump.
- Copy lives in both `messages/en-US.json` and `messages/zh-TW.json`.
- `DESIGN.md` gains a **Public Surface Exception** next to the Native Scale Rule, so landing copy on `/login` isn't read as a violation of "no marketing typography in product surfaces".

## Proxy fix worth reviewing

`public/*` assets are **not** excluded from the proxy matcher, so `GET /readme-hero.jpg` returned `302 → /login` for anonymous visitors — the image was broken for exactly the audience this page targets. The asset is renamed `readme-hero.jpg → hero.jpg` (shared with both READMEs, both updated) and added to the matcher's exclusion list, alongside `sw.js` / `manifest.webmanifest`. `tests/unit/proxy.test.ts` covers it.

## Follow-up, not fixed here

`hero.jpg` still carries the **old brand** ("Assets Tracker"), which now sits next to the `astt` card on the same screen. It is the same image already at the top of both READMEs, so this PR doesn't introduce the inconsistency — but the landing page is the place it looks worst. Replacing the asset is a drop-in (same filename, same dimensions), or delete the `` block in `src/app/login/page.tsx` if you'd rather ship without a visual.

## Verified

`pnpm format:check`, `pnpm lint`, `pnpm typecheck`, `pnpm build`, `pnpm test:unit` (1003 passing). Rendered locally at desktop + mobile, light and dark; the hero image loads anonymously (`200`, was `302`).

https://claude.ai/code/session_01L8h3gpHC9CoPsDp6FSHccw


mike840609:
Superseded by #716. With a real landing page at `/`, the pitch on `/login` would repeat the same headline one click later, so `/login` goes back to being a sign-in card. #716 contains this branch's commits and reverts the login-only parts; the DESIGN.md exception survives there, reworded for the landing page.

PR #717, opened 2026-08-25
perf: reuse history fill for snapshot reconciliation

mike840609:
## Summary
- `getSnapshotReconciliationWarning` now accepts pre-computed normalized snapshots so the History page doesn't re-scan snapshots/accounts/FX a second time.
- History page wires the reconciliation call off the same history fill instead of issuing an independent query set.

## Testing
- `pnpm exec tsc --noEmit`
- `pnpm exec vitest run tests/unit/history-service.test.ts tests/unit/public-demo-route-policy.test.ts`
- `pnpm exec prettier --check` + `pnpm exec eslint` on touched files

### DIFF
--- a/src/app/(main)/history/page.tsx
+++ b/src/app/(main)/history/page.tsx
@@ -7,7 +7,7 @@ import { HistoryPullRefresh } from "@/components/history/history-pull-refresh";
 import { HistoryView } from "@/components/history/history-view";
 import {
   getFullNormalizedHistory,
-  getSnapshotReconciliationWarning,
+  getSnapshotReconciliationWarningFromHistory,
   hasForeignCurrencySnapshots,
 } from "@/lib/services/history-service";
 import { countActiveAccounts } from "@/lib/services/account-service";
@@ -20,13 +20,11 @@ async function HistoryContent() {
   const userId = session.user.id;
   const settingsP = getOrCreateSettings(userId);
   const baseCurrencyP = settingsP.then((s) => s.baseCurrency);
-  const snapshotsP = Promise.all([baseCurrencyP, settingsP]).then(([currency]) =>
-    getFullNormalizedHistory(userId, currency),
+  const snapshotsP = baseCurrencyP.then((currency) => getFullNormalizedHistory(userId, currency));
+  // Reuses the history fill instead of re-reading snapshots/accounts/rates.
+  const reconciliationP = Promise.all([snapshotsP, baseCurrencyP]).then(([snapshots, currency]) =>
+    getSnapshotReconciliationWarningFromHistory(userId, currency, snapshots),
   );
-  const reconciliationP: ReturnType<typeof getSnapshotReconciliationWarning> = Promise.all([
-    snapshotsP,
-    baseCurrencyP,
-  ]).then(([snapshots, currency]) => getSnapshotReconciliationWarning(userId, currency, snapshots));
 
   const [allMessages, snapshots, settings, accountCount, reconciliationWarning, converted] =
     await Promise.all([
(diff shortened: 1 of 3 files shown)


mike840609 (commented):
## Review — 3 issues, one of them likely defeats the PR's goal

### 1. Passing the array into a `"use cache"` function makes the cache key the entire history 🔴

`getSnapshotReconciliationWarning` still starts with `"use cache"`, and the new parameter is now part of its cache key:

```ts
export async function getSnapshotReconciliationWarning(
  userId: string,
  targetBaseCurrency: string,
  normalizedSnapshots?: NormalizedSnapshot[],   // ← becomes part of the cache key
): Promise<SnapshotReconciliationWarning | null> {
  "use cache";
```

Every call now has to serialize the user's full normalized history to build the key. Three consequences, all working against the stated goal:

- **Key size grows with history length** — a user with years of snapshots serializes that entire array on every request just to look up the cache.
- **Cache reuse collapses** — the key changes whenever any snapshot value changes, so entries are effectively single-use rather than being reused across the `cacheLife("minutes")` window.
- **Memory** — each distinct history produces its own cache entry keyed by the full payload.

The saved work is 3 queries (`findFirst` + `account.findMany` + `getAllExchangeRates`, the last of which is itself cached). It's very plausible this trade is net negative.

Data shouldn't cross a `"use cache"` boundary as an argument. Two ways out:

- Keep the cached function arg-light (`userId`, `targetBaseCurrency`) and extract the comparison into an **uncached** inner function that takes `normalizedSnapshots`, letting the History page call that directly; or
- Drop the parameter and accept the duplicate reads — they're cached anyway, so the original cost may be smaller than it looks.

### 2. The new guard is dead code, and its falsiness check is wrong 🔴

```ts
const currentSummary = await getCachedNetWorthSummary(userId, targetBaseCurrency);
if (!snapshotNetWorth && !currentSummary) return null;
const currentNetWorth = currentSummary.netWorth;
```

`getCachedNetWorthSummary` returns an object, so `!currentSummary` is always `false` and the condition can never be true — the line never fires. It also reads as a null-guard for `currentSummary` while providing no actual protection for the `.netWorth` access on the next line.

Separately, `!snapshotNetWorth` is true when `snapshotNetWorth === 0`, which is a legitimate net worth, not a missing value. Worth either deleting the line or writing the check that was actually intended.

### 3. The two branches disagree on which snapshot is "latest" 🟡

They use different tie-breaking rules for same-date snapshots:

- **Array path** takes the last element, which `normalizeSnapshots` selected via `isBetterDuplicate` — that prefers `matchesTarget` (stored `baseCurrency === targetBaseCurrency`) **first**, and only then `createdAt`.
- **DB path** orders by `[{ date: "desc" }, { createdAt: "desc" }]` — newest `createdAt` wins regardless of stored currency.

So for a user with two snapshots on the same latest date — say a USD one created at 09:00 and a EUR one created at 14:00, with `targetBaseCurrency = USD` — the array path picks the USD row and the DB path picks the EUR row (then converts). Different `snapshotNetWorth`, so potentially a different reconciliation warning depending on which branch ran.

The array path's rule is arguably the better one, but this is an unannounced behavior change and the two branches inside one function now disagree. Worth aligning them and saying so in the description.

### Minor

**The empty case falls into the slow path.** `if (normalizedSnapshots && normalizedSnapshots.length > 0)` sends a user with *no* snapshots into the `else` branch, running the exact three queries this PR set out to avoid — only to hit `if (!latestSnapshotRow) return null`. Branching on `normalizedSnapshots !== undefined` and returning `null` for an empty array short-circuits the cheapest case.

**Redundant `Promise.all`:**

```ts
const snapshotsP = Promise.all([baseCurrencyP, settingsP]).then(([currency]) =>
  getFullNormalizedHistory(userId, currency),
);
```

`settingsP` is discarded, and `baseCurrencyP` already derives from it so it can't resolve earlier. `baseCurrencyP.then((currency) => getFullNormalizedHistory(userId, currency))` is equivalent.

---

PR #718, opened 2026-08-25
perf: index snapshots by user and base currency

mike840609:
## Summary
- Adds `@@index([userId, baseCurrency])` to `NetWorthSnapshot` so `hasForeignCurrencySnapshots` (and any future per-currency snapshot query) doesn't scan the user's full history.
- Ships the matching Prisma migration.

## Testing
- `pnpm exec tsc --noEmit`
- Migration SQL reviewed; no runtime code change.

### DIFF
--- a/prisma/migrations/20260825000000_add_snapshot_currency_index/migration.sql
+++ b/prisma/migrations/20260825000000_add_snapshot_currency_index/migration.sql
@@ -1,0 +1,2 @@
+-- CreateIndex
+CREATE INDEX IF NOT EXISTS "NetWorthSnapshot_userId_baseCurrency_idx" ON "NetWorthSnapshot"("userId", "baseCurrency");
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -274,6 +274,7 @@ model NetWorthSnapshot {
   createdAt        DateTime @default(now()) @db.Timestamptz(3)
 
   @@unique([userId, date, baseCurrency])
+  @@index([userId, baseCurrency])
   @@index([userId, date(sort: Desc)])
 }
 
--- a/src/lib/services/history-service.ts
+++ b/src/lib/services/history-service.ts
@@ -643,9 +643,25 @@ export async function hasForeignCurrencySnapshots(
   cacheTag(`history:${userId}`);
   cacheLife("minutes");
 
-  const row = await prisma.netWorthSnapshot.findFirst({
-    where: { userId, baseCurrency: { not: targetBaseCurrency } },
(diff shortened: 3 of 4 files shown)


mike840609 (commented):
## Review — the index is fine, but the query can't fully exploit it

### The `!=` predicate can't seek 🟡

The only consumer named in the description filters with a **not-equal**:

```ts
const row = await prisma.netWorthSnapshot.findFirst({
  where: { userId, baseCurrency: { not: targetBaseCurrency } },
  select: { id: true },
});
```

A btree on `(userId, baseCurrency)` can only seek on the `userId` prefix here — `baseCurrency != ?` isn't a seekable condition, so Postgres scans every index entry under that `userId` and filters. The index is still a real win (narrow, index-only, no heap access, versus reaching through `(userId, date DESC)` and fetching `baseCurrency` from the heap), so I'd keep it. But "doesn't scan the user's full history" only holds in the *positive* case, where a differing currency is found early. In the common case — every snapshot already in the base currency, returning `false` — it still has to traverse the user's entire prefix to prove the negative.

Since entries are sorted by `baseCurrency` within each `userId`, the existence check collapses to two index seeks:

```ts
const [lo, hi] = await Promise.all([
  prisma.netWorthSnapshot.findFirst({
    where: { userId },
    select: { baseCurrency: true },
    orderBy: { baseCurrency: "asc" },
  }),
  prisma.netWorthSnapshot.findFirst({
    where: { userId },
    select: { baseCurrency: true },
    orderBy: { baseCurrency: "desc" },
  }),
]);
if (!lo) return false;
return lo.baseCurrency !== targetBaseCurrency || hi!.baseCurrency !== targetBaseCurrency;
```

If both the smallest and largest `baseCurrency` for the user equal the target, every value in between must too. That's O(1) regardless of history size, and it's what makes this new index actually pay for itself. As it stands the PR adds the index but leaves the query unable to use it beyond the prefix — worth landing both together, otherwise it's write-amplification for a partial gain.

### Migration note 🟢

`CREATE INDEX IF NOT EXISTS` (rather than `CONCURRENTLY`) takes a lock that blocks writes to `NetWorthSnapshot` while the index builds. Prisma wraps migrations in a transaction and `CONCURRENTLY` can't run inside one, so this is the normal trade-off and I wouldn't change it — just worth knowing the write-blocking window scales with table size on the production deploy.

The `IF NOT EXISTS` guard is consistent with the repo's existing migration style, and the schema change itself is correct — `@@unique([userId, date, baseCurrency])` can't serve this query since `date` sits between the two columns being filtered.

---

PR #719, opened 2026-08-25
perf: scope FX banner invalidation to the user

mike840609:
## Summary
- `getUnresolvedRatePairs` now only carries user-scoped tags (`accounts:{userId}`, `goals:{userId}`, `history:{userId}`) plus the shared `exchange-rates` tag.
- Removes global `goals` / `snapshots` tags so one user's goal edit or any cron snapshot no longer rebuilds every user's FX warning banner.

## Testing
- `pnpm exec tsc --noEmit`
- `pnpm exec vitest run tests/unit/exchange-rate-service.test.ts tests/unit/public-demo-route-policy.test.ts`
- `pnpm exec prettier --check` + `pnpm exec eslint` on touched file

### DIFF
--- a/src/lib/services/exchange-rate-service.ts
+++ b/src/lib/services/exchange-rate-service.ts
@@ -461,9 +461,7 @@ export async function getUnresolvedRatePairs(
   "use cache";
   cacheTag("exchange-rates");
   cacheTag(`accounts:${userId}`);
-  cacheTag("goals");
   cacheTag(`goals:${userId}`);
-  cacheTag("snapshots");
   cacheTag(`history:${userId}`);
   cacheLife("minutes");
 


mike840609 (commented):
## Review — safe, verified against every write path ✅

The removal is only correct if every writer that revalidates the global `snapshots` / `goals` tags *also* revalidates a user-scoped tag this function still carries. I traced all of them:

**Snapshot writers** — `getUnresolvedRatePairs` does genuinely read snapshots (`distinct baseCurrency`), so the tag isn't semantically irrelevant, just redundant:

| Writer | Tags revalidated | Covered by remaining tags? |
|---|---|---|
| `api/cron/snapshot` | `snapshots` **+ `history:${userId}` per successful user** | ✅ |
| `api/snapshots/[id]` | `invalidateScopedTag({ globalTag: "snapshots", userTag: \`history:${userId}\` })` | ✅ |
| `api/settings/data` (import) | `snapshots` + `history:${userId}` | ✅ |
| `snapshot-service.createSnapshot` | none of its own — only called by the cron above | ✅ |

The cron was the one I most expected to be a problem, since `revalidateTag("snapshots", "max")` is global. But the loop right beneath it covers each user individually, so dropping the global tag changes nothing:

```ts
if (snapshots.length > 0) revalidateTag("snapshots", "max");
for (const userId of successfulUserIds) {
  revalidateTag(`history:${userId}`, "max");
}
```

**Goal writers** — `invalidateImportCaches` revalidates both `goals` and `goals:${userId}`, and `invalidateScopedTag` always revalidates the user tag (the global one only for `formal` principals). So `goals:${userId}` is never missed.

The invalidation reasoning holds, and the win is real: a cron run over N users no longer busts every user's FX banner.

### One pre-existing gap, unrelated to this change

Since this PR is curating the tag list, it's a natural moment to raise it — this is **not** a regression introduced here. `getUnresolvedRatePairs` reads price-cache currencies:

```ts
const prices = await prisma.priceCache.findMany({
  where: { symbol: { in: symbols } },
  select: { currency: true },
  distinct: ["currency"],
});
for (const price of prices) currencies.add(price.currency);
```

…but carries no `prices` cacheTag, while four other sites do `revalidateTag("prices", ...)`. If a price refresh changes the currency a symbol is denominated in, the banner won't recompute until `cacheLife("minutes")` expires. Bounded and low-severity, but adding `cacheTag("prices")` would make the tag set match what the function actually reads.

---

PR #720, opened 2026-08-25
perf: memoize exchange-rate lookups in net worth

mike840609:
## Summary
- `computeNetWorthSummary` now memoizes resolved (from→to) pairs so repeated holding conversions hit a map lookup instead of the full direct/inverse/USD-cross chain every time.
- No behavior change; same unresolved-pair warning path.

## Testing
- `pnpm exec tsc --noEmit`
- `pnpm exec vitest run tests/unit/net-worth-service.test.ts`
- `pnpm exec prettier --check` + `pnpm exec eslint` on touched file

### DIFF
--- a/src/lib/services/net-worth-service.ts
+++ b/src/lib/services/net-worth-service.ts
@@ -225,6 +225,8 @@ export async function computeNetWorthSummary(
   // not once per holding.
   const rateMemo = new Map<string, number>();
   function getRate(from: string, to: string): number {
+    // Deliberate fast path: resolveRate opens with the same identity check, so
+    // this only saves the call — keep the two in step if either changes.
     if (from === to) return 1;
     const memoKey = `${from}_${to}`;
     const memoized = rateMemo.get(memoKey);
@@ -238,6 +240,12 @@ export async function computeNetWorthSummary(
       warnedPairs.add(memoKey);
       log.warn("rates.unresolved", { from, to, userId, baseCurrency });
     }
+    // Memoize the fallback too: a miss is the most expensive path (resolveRate
+    // exhausted direct, inverse and the USD cross before giving up), and
+    // allRatesMap is fixed for this call, so the answer cannot change. The
+    // warning is already deduped by warnedPairs and fires on the first miss,
+    // above, so short-circuiting here suppresses nothing.
+    rateMemo.set(memoKey, 1);
     return 1;
   }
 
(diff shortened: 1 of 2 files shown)


mike840609 (commented):
## Review — correct, but it misses the case it most wants to optimize

### Unresolved pairs are never memoized 🟡

```ts
const resolved = resolveRate(allRatesMap, from, to);
if (resolved !== undefined) {
  rateMemo.set(memoKey, resolved);
  return resolved;
}
if (!warnedPairs.has(memoKey)) { ... }
return 1;    // ← nothing cached
```

Only the hit path populates the memo. When `resolveRate` returns `undefined`, the next holding with the same pair runs the full chain again — and that's the *most* expensive path, since a miss means it already exhausted all three lookups:

```ts
if (from === to) return 1;
const direct = rateMap.get(`${from}_${to}`);        // miss
if (direct !== undefined) return direct;
const inverse = rateMap.get(`${to}_${from}`);       // miss
if (inverse !== undefined && inverse !== 0) return 1 / inverse;
return crossRateViaUsd((key) => rateMap.get(key), from, to);   // miss, several more lookups
```

So a portfolio holding 50 positions in an unpriced currency still pays the full resolution chain 50 times — exactly the "once per holding, not once per currency combination" cost the PR describes eliminating. Caching the fallback fixes it:

```ts
if (resolved !== undefined) {
  rateMemo.set(memoKey, resolved);
  return resolved;
}
if (!warnedPairs.has(memoKey)) {
  warnedPairs.add(memoKey);
  log.warn("rates.unresolved", { from, to, userId, baseCurrency });
}
rateMemo.set(memoKey, 1);
return 1;
```

This is behavior-preserving: `allRatesMap` is fixed for the lifetime of the call so `resolveRate` is deterministic here, the function already returns `1` for misses, and the `log.warn` is separately deduped by `warnedPairs` — so short-circuiting on the memo can't suppress a warning that would otherwise fire. `warnedPairs` is local to `computeNetWorthSummary` and only used for that dedup, so nothing downstream depends on it being consulted every call.

### Minor 🟢

`if (from === to) return 1;` duplicates a check `resolveRate` already does on its first line. Harmless and it does save a call, but it puts the identity invariant in two places — if `resolveRate` ever grows a same-currency nuance, this shadow copy won't follow. Worth a short comment noting it's a deliberate fast path.

The memo key format matches the existing `warnedPairs` key, and ISO currency codes can't contain `_`, so there's no collision risk between pairs.

---

PR #730, opened 2026-09-01
fix(recurring): re-anchor nextRunDate when a paused rule is resumed

mike840609:
## Problem

Pausing a recurring cash or DCA rule only flips `isActive`; `nextRunDate` stays frozen at the first occurrence missed while paused. Resuming hands that stale date to the catch-up loop in `computeDueOccurrences`, which replays **every occurrence in the paused window**:

- cash rule: N deposits/withdrawals land on the balance at once
- DCA rule: N buys, all priced at today's price, and cash debited N times

Verified with the pure helper: a monthly rule with `nextRunDate` 2026-05-01 resumed on 2026-09-02 yields five occurrences.

## Fix

Resuming (`isActive: false → true`) now re-anchors `nextRunDate` to the first scheduled occurrence on/after today, reusing `firstOccurrenceOnOrAfter` — the same clamp a `startDate` edit already applies in both PATCH routes. A no-op `isActive: true` on an already-active rule is left untouched so a pending occurrence the cron has not posted yet is not skipped.

## Tests

- new: resume re-anchors for cash and investment rules, and schedules nothing when no occurrence is due today
- new: no-op `isActive: true` on an active rule does not touch `nextRunDate`
- updated: the "became due on resume" tests now pin the clock to a scheduled day, so they encode "only today's occurrence is due" rather than relying on the replay
- `pnpm test:unit` 1079/1079, `format:check` / `lint` / `typecheck` clean

Surfaced by a data-correctness audit; the remaining findings are filed as separate issues.


mike840609:
Validated the resume semantics. Re-anchoring only on the paused → active transition avoids replaying the paused window, while leaving an already-active `isActive: true` PATCH untouched preserves genuinely pending occurrences after a failed cron. The updated tests cover both cash and DCA paths plus the no-op active case. One minor edge case to consider later: resuming a rule whose `endDate` is already in the past may briefly return active until materialization deactivates it, but this is not a data-correctness blocker. Merge-ready.

PR #737, opened 2026-09-01
fix(recurring): skip or refetch stale DCA prices

### LINKED ISSUE #731: DCA materialization buys at a stale PriceCache price after a partial price-refresh failure
## Where

`src/lib/services/recurring-investment-service.ts:48-54` (`resolvePrice`) and `src/app/api/cron/snapshot/route.ts:181-185`.

## Problem

`resolvePrice` accepts any `PriceCache` row with `price > 0`; it never looks at `updatedAt`. The cron only aborts when the price refresh is a `total_failure`; a `partial_success` (one symbol rate-limited or delisted) continues into `materializeDueInvestments`. A due DCA rule for that symbol is then converted to shares at whatever price is left in the cache — possibly days old — and the BUY row, holding quantity, and cash debit are all persisted with that wrong share count.

## Expected

Skip the rule (leave `nextRunDate` untouched, same as the "no price" and "no rate" branches) when the cached price is older than a freshness threshold, e.g. `SNAPSHOT_STALE_MS` / 48 h from `src/lib/refresh-policy.ts`, or re-fetch the symbol before pricing the buy.

## Severity

Medium — silent, permanent, and only surfaces on days the provider drops a symbol.

Found during a data-correctness audit (see PR "fix(recurring): re-anchor nextRunDate when a paused rule is resumed" for the high-severity sibling).

mike840609:
## Problem

`resolvePrice` in `src/lib/services/recurring-investment-service.ts` accepted any `PriceCache` row with `price > 0` and never looked at `updatedAt`.

The daily cron refreshes prices immediately before calling `materializeDueInvestments`, but it only aborts on a `total_failure`. A `partial_success` (one symbol rate-limited or delisted) continues, so a due DCA rule for that symbol was converted to shares at whatever price was left in the cache — possibly days old. The BUY row, the holding quantity increment and the cash debit were all persisted with that wrong share count: silent, permanent, and only visible on days a provider drops a symbol.

## Fix

- `src/lib/refresh-policy.ts`: new `DCA_PRICE_MAX_AGE_MS = 48h`. The cron refreshes held symbols right before materializing, so ~24h is the normal age; 48h means that symbol's refresh failed at least once.
- `src/lib/services/recurring-investment-service.ts`: `resolvePrice` also selects `updatedAt` and takes the run clock (`now`, threaded from the `now` the service already receives, so tests can pin it). A cached row older than the threshold is logged as `log.warn("cron.investment.stale_price", { symbol, ageMs })` and treated like a missing row — it falls through to the existing live-fetch branch (`fetchStockPrices` / `fetchCryptoPrices`), which re-caches on success. If that fetch fails or returns nothing, `resolvePrice` returns `null` and the caller's existing `cron.investment.skip_no_price` branch skips the rule with `nextRunDate` untouched, so it retries on the next run.
- The cron route is unchanged.

## Tests

`tests/unit/recurring-investment-service.test.ts` — the mocked `priceCache.findUnique` now returns `updatedAt`, and `log.warn` is captured. Three new cases:

1. stale cached row + successful live refetch → the BUY uses the refetched price (250, so 4 shares instead of 5), the cache is upserted, and `cron.investment.stale_price` is logged.
2. stale cached row + empty refetch → rule skipped, no transaction/holding/account writes, `nextRunDate` not advanced, both `stale_price` and `skip_no_price` logged.
3. cached row 1s inside the window → unchanged behaviour, no refetch, no warning (regression guard).

`pnpm test:unit` (1079 tests), `pnpm format:check`, `pnpm lint`, `pnpm typecheck` all clean.

## Decisions

- **48h, matching `FX_RATES_STALE_MS` / `SNAPSHOT_STALE_MS`.** A tighter bound (e.g. 25h) would refetch on every ordinary cron jitter; 48h only trips after a genuine missed refresh, which is exactly the partial-failure case in the issue.
- **Refetch first, skip second** (the issue offered either). Refetching keeps the rule on schedule whenever the symbol is merely transiently unavailable, and only falls back to the skip-and-retry path — which is already the behaviour for "no price" and "no rate" — when it truly can't be priced. No new failure mode: the fetch is the same call the brand-new-symbol path already made.
- **`ageMs` is computed against the passed `now`, not wall-clock.** That keeps the check consistent with the Taiwan-business-day cutoff the same call already uses and makes it deterministic under test. A negative age (clock skew, or an integration test running a past `RUN_DATE` against a just-written cache row) counts as fresh.
- The two `/api/accounts/[id]/recurring-investments` routes that materialize on create/resume call this with `new Date()`, so a rule created against a >48h-old cache row now triggers one live fetch instead of pricing off the stale row — the correct trade for a user-initiated action.

Fixes #731

### DIFF
--- a/src/lib/refresh-policy.ts
+++ b/src/lib/refresh-policy.ts
@@ -29,5 +29,14 @@ export const FX_RATES_STALE_MS = 48 * 60 * 60 * 1000;
  */
 export const SNAPSHOT_STALE_MS = 48 * 60 * 60 * 1000;
 
+/**
+ * Refuse to price a recurring-investment (DCA) buy from a cached price older
+ * than this. The daily cron refreshes held symbols immediately before
+ * materializing, so ~24h is the normal age; 48h means that symbol's refresh
+ * failed at least once (rate limit, delisting) and the cached price can no
+ * longer stand in for today's.
+ */
+export const DCA_PRICE_MAX_AGE_MS = 48 * 60 * 60 * 1000;
+
 /** Minimum client-side cooldown after any manual refresh attempt. */
 export const CLIENT_REFRESH_COOLDOWN_MS = 15_000;
--- a/src/lib/services/recurring-investment-service.ts
+++ b/src/lib/services/recurring-investment-service.ts
@@ -6,6 +6,7 @@ import { computeDueOccurrences, utcDateOnly } from "./recurring-cash-service";
 import { taiwanCalendarDay } from "@/lib/app-day";
 import { getFreshExchangeRates, resolveRate } from "./exchange-rate-service";
 import { fetchStockPrices, fetchCryptoPrices } from "./price-service";
+import { DCA_PRICE_MAX_AGE_MS } from "@/lib/refresh-policy";
 
 class HoldingCurrencyMismatchError extends Error {
   constructor(readonly holdingCurrency: string) {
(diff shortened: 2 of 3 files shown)


mike840609:
Validated the stale-price handling. The important correctness property is preserved: once a cached DCA price is outside the freshness window, the service refetches; if that cannot produce a usable price, it skips without advancing `nextRunDate`, so the occurrence is retried instead of being silently consumed. The run clock is threaded through the age check, which also makes this deterministic under tests. CI/E2E are green. Merge-ready.

PR #739, opened 2026-09-01
fix(cron): settle expired options at intrinsic value, not a stale premium

### LINKED ISSUE #732: Expired-option sweep writes the whole position off as worthless
## Where

`src/app/api/cron/snapshot/route.ts:87-118`.

## Problem

Contracts whose `expiration` is before the Taiwan business day are zeroed with a `SELL` row (`note: "Expired"`, no `unitPrice`). That is correct for out-of-the-money contracts, but an in-the-money contract has intrinsic value at expiry: it is exercised or assigned, not vaporised. Today the sweep makes that intrinsic value disappear from net worth overnight with no corresponding cash or underlying movement, and the cost-basis path treats the null-priced SELL as a disposal at average cost.

## Expected (pick one)

- price the `SELL` at the last cached price (or intrinsic value from `strike` vs the underlying's cached price) so cost basis records the realised P&L, and credit the account's cash with that value; or
- leave ITM contracts untouched and surface them for manual close, sweeping only worthless (OTM / no price) contracts automatically.

Either way the current behaviour should be documented next to the sweep.

## Severity

Medium — silent net-worth change without user action; affects anyone holding options through expiry.

mike840609:
> **Reworked after review — this supersedes the "sold at the last cached close" approach.** The review was right: the option's own `PriceCache` row is not a trustworthy settlement source. Settlement is now derived from the expiration-day underlying close and intrinsic value, and the sweep refuses to settle at all when that price cannot be established.

## Problem

The daily snapshot cron's step 0 sweeps every option contract whose `expiration` precedes the Taiwan business day. On `master` it zeroes the quantity and writes a `SELL` with `note: "Expired"` and **no** `unitPrice` — correct for an out-of-the-money contract, wrong for an in-the-money one, which is exercised or assigned at expiry rather than vaporised (#732).

The first attempt at this PR closed each contract at its last cached option premium. That is unsafe. `vercel.json` runs the cron at `21:30 UTC` and the sweep intentionally runs **before** `refreshAllPrices()`, so for a Friday-expiring US contract the Saturday-Taipei run reads the `PriceCache` row left by the *previous* cycle — Thursday's premium. A contract worth $1 on Thursday that expired worthless on Friday would have been auto-closed at $1 and credited $100/contract of cash that never existed.

## Fix

`src/app/api/cron/snapshot/route.ts` — settlement value is now the contract's **intrinsic value on its expiration day**, computed from a live quote for the **underlying**:

- One `fetchStockPrices([...distinct underlyingSymbols])` call (live, not `PriceCache`) plus one `getFreshExchangeRates()` read, both issued once before the loop and only when there is something to sweep, so the existing "FX map loads once per run" bound is unchanged. The option's own cached premium is never read — `prisma.priceCache` is gone from this path entirely.
- Intrinsic per share = `max(spot - strike, 0)` for a `CALL`, `max(strike - spot, 0)` for a `PUT`, in the option's currency.
- **Intrinsic > 0** → zero the quantity, `SELL` with `unitPrice` = the per-share intrinsic value (same convention as manual option transactions; `investment-cost-basis-service` applies the contract multiplier itself), and `cashBalance += intrinsic x quantity x (contractMultiplier ?? 100)`, converted into the account's currency via `resolveRate`.
- **Intrinsic = 0** (at or out of the money) → `master`'s exact behaviour: zero the quantity, `SELL` with `note: "Expired"` and a **null** `unitPrice`, no cash. `unitPrice` is never written as `0` — `importHoldingTransactionUnitPrice` rejects a non-null `unitPrice <= 0`, so a stored `0` would make backups unimportable.
- Arithmetic stays in `Decimal` and is snapped to 8 decimal places before it reaches Prisma (the `toDbMoneyDelta` invariant). `1.07 x 3 x 100` writes `321`, not float's `321.00000000000006`.
- Everything for one holding stays inside the existing per-holding `$transaction`, gated on the same optimistic `quantity` guard; cash moves only when the guarded zeroing reported `count === 1`.
- The sweep stays exactly where it was in the pipeline. It no longer depends on `PriceCache` at all, so its position is now purely about ordering relative to the snapshot write — the "read the close before the refresh overwrites it" reasoning is deleted from the comment block and the docs.

## Trustworthiness gate

Automatic settlement runs only when `businessDay - expiration === exactly one day` (both are UTC-midnight dates, compared on `getTime()` against `24 x 60 x 60 x 1000`).

`taiwanCalendarDay(cron at 21:30 UTC on day D) === D+1`, and that run happens after day D's US close (20:00 UTC EDT / 21:00 UTC EST). A diff of exactly one day therefore means this run is the **first sweep after expiry** and the live quote *is* the expiration-day close. A larger diff means a cron cycle was missed and the underlying has traded since — the expiration-day price can no longer be established.

## Deferral behaviour (deliberate change from `master`)

When the expiration-day price cannot be trusted, the holding is left **completely untouched** — no zeroing, no `SELL`, no cash — and a `log.warn("cron.options.expire_deferred", { symbol, reason })` records why. The position stays visible in the UI for the user to close manually, which is the "leave settlement pending/manual" branch the review asked for. `master` always zeroed; that silent loss of value is what this replaces.

Reasons emitted:

| `reason` | Condition |
| --- | --- |
| `expiry_window` | `businessDay - expiration !== 1 day` (a cron cycle was missed) |
| `missing_option_terms` | No `underlyingSymbol` / `optionType` / `strike` — imported rows can lack the OCC fields the POST route derives, and intrinsic value is undefined without them |
| `underlying_unavailable` | The live fetch returned no quote for the underlying (a provider outage also logs `cron.options.expire_quotes_failed` and defers every contract instead of aborting the run) |
| `underlying_currency_mismatch` | The quote's currency differs from the option holding's `currency`; spot and strike must be the same unit before they are subtracted |
| `rate_unresolved` | No direct, inverse, or USD-cross rate from the option's currency to the account's |

## Decisions (please veto if you disagree)

- **Cash, not shares.** Auto-exercise into the underlying is still deliberately *not* modelled: it would mint share lots the user never confirmed, at a cost basis nobody chose. Cash equal to the intrinsic value is the closest faithful summary.
- **FX is resolved before the ITM/OTM split, so an unresolvable pair defers even a worthless contract.** One code path, and the sweep never mutates a holding whose value it cannot express in the account's currency. The alternative (settle worthless anyway, since 0 needs no conversion) would close the position while the account's FX configuration is visibly broken.
- **Missing OCC terms defer rather than expire worthless.** Writing a legacy/imported contract off is the same silent value loss the review objected to, just in the safe direction; leaving it visible lets the user decide.
- **The gate is calibrated to US market hours,** which is where the OCC contracts this feature models trade. A contract on another exchange is settled on the same one-day rule; if that proves wrong for a real instance the gate is one constant.
- **A stuck contract warns on every run.** That is intentional: the operator sees it in the logs, and the user sees the still-open position.
- Untouched: `src/lib/changelog.ts` and the `package.json` version.

## Docs

`docs/ARCHITECTURE.md` → "Expired options" rewritten for intrinsic-value settlement and the deferral rule; the "last cached close" reasoning is deleted. The comment block above the sweep carries the same reasoning plus the stale-premium hazard that motivated it.

## Tests

`tests/unit/cron-snapshot-route.test.ts` — the `priceCache`-driven mocks are replaced with a `fetchStockPrices` mock feeding live underlying quotes. The default fixture contract expires exactly one business day before the cutoff and is at the money, so every pre-existing assertion in the file still describes the worthless path.

1. ITM **call** → `SELL` carries the per-share intrinsic value (`101.07 - 100 = 1.07`) and cash is credited `1.07 x 3 x 100 = 321` exactly (also exercises the `contractMultiplier == null` → 100 fallback), and `priceCache.findMany` is asserted **not** to have been called.
2. ITM **put** → `max(500 - 497.5, 0) = 2.5` per share, `500` credited.
3. OTM call, OTM put, and an at-the-money call → all three written off worthless: `SELL` with no `unitPrice`, no `account.update`. Regression guard for `master`'s behaviour.
4. **The review fix, encoded:** option premium `1.00` sitting in `PriceCache` but the underlying closed at `98` against a `100` strike → **zero** cash, `SELL` with no `unitPrice`, `priceCache` never read.
5. `businessDay - expiration > 1 day` → deferred, holding untouched, `reason: "expiry_window"`.
6. Underlying quote missing → deferred (`underlying_unavailable`).
7. Live quote fetch throws → deferred, `cron.options.expire_quotes_failed` warned, and the rest of the cron still completes.
8. Cross-currency underlying quote → deferred (`underlying_currency_mismatch`).
9. Unresolvable FX pair to the account → deferred (`rate_unresolved`).
10. Imported contract with no OCC terms → deferred (`missing_option_terms`).
11. Optimistic quantity guard fails mid-sweep (concurrent trade) on an ITM contract → no `SELL`, no cash.
12. Non-USD account → `1.07 x 3 x 100 x 31.5 = 10111.5` credited in TWD.
13. The sweep's quote fetch precedes `refreshAllPrices`, pinning its position in the pipeline.

Every deferral assertion goes through one helper that checks `holding.updateMany`, `holdingTransaction.create`, and `account.update` were all skipped, so "untouched" means untouched.

`pnpm test:unit` (119 files / 1113 tests), `pnpm format:check`, `pnpm lint`, `pnpm typecheck` all clean.

Fixes #732


mike840609:
Blocking correctness concern: the cached option price used here is not reliably the expiration-day close under the current cron schedule/order.

`vercel.json` runs `/api/cron/snapshot` at `21:30 UTC` (05:30 Taipei). The expiry sweep intentionally runs *before* `refreshAllPrices()`. For a Friday-expiring US option, the Saturday-morning Taipei sweep therefore commonly sees the PriceCache value left from the prior cron cycle, i.e. potentially Thursday's option price, not Friday's expiration close. An option that was worth $1 on Thursday but expired worthless Friday could then be auto-closed at $1 and credit $100/contract of cash that never existed.

I don't think stale option premium is safe as the settlement source. Prefer deriving expiry value from expiration-day underlying close and intrinsic value (`max(S-K,0)` for calls / `max(K-S,0)` for puts) if that data is available, or leave settlement pending/manual when a trustworthy expiration-day price cannot be established. Until this is addressed, I would not merge this PR despite green CI.

PR #740, opened 2026-09-01
fix(db): widen money columns to Decimal(28,8) and bound imports

### LINKED ISSUE #735: Decimal(18,8) caps every amount and balance below 10^10 — reachable for IDR/VND/KRW users
## Where

`prisma/schema.prisma` (`cashBalance`, `amount`, `quantity`, `unitPrice`, `targetAmount`, `rate` are all `Decimal(18,8)`), `src/lib/validators.ts:18-22` (`CRUD_DECIMAL_ABS_MAX = 10_000_000_000`), `src/lib/validators.ts:429` (import `decimalSchema` has no magnitude bound).

## Problem

`numeric(18,8)` stores at most 9,999,999,999.99999999. The CRUD validators reject a single input ≥ 1e10, but:

- `cashBalance` is updated with `increment` / `decrement`, so a sequence of valid deposits can push the stored balance past the ceiling → Postgres `numeric field overflow` → 500 from the route, mid-transaction.
- the import path validates with `decimalSchema` (`z.number().finite()` / decimal string) and no upper bound, so a backup containing such a balance fails the whole import with a generic 500.

1e10 IDR ≈ US$600k, 1e10 VND ≈ US$400k, 1e10 KRW ≈ US$7M — plausible net-cash figures for the open-source audience, even if TWD users never get near it.

## Expected

Either widen the columns (e.g. `Decimal(28,8)`; `NetWorthSnapshot` already uses `Decimal(18,2)` = 10^16) and bump the validator cap to match, or enforce the ceiling in the import validator and turn the overflow into a 400 with a field path. A one-off migration is enough; no data conversion is needed.

## Severity

Medium — hard 500, not silent corruption, but data-blocking for the affected currencies.

mike840609:
## Problem

`numeric(18,8)` stores at most `9,999,999,999.99999999`, and every money/quantity column used that width.

- `CRUD_DECIMAL_ABS_MAX = 10_000_000_000` rejected a *single* input ≥ 1e10, but `cashBalance` is updated with `increment` / `decrement`. A sequence of individually valid deposits could push the stored balance past the ceiling → Postgres `numeric field overflow` → 500 from the route, mid-transaction.
- The import path validated amounts with an unbounded `decimalSchema` (`z.number().finite()` / decimal string), so a backup containing such a balance failed the whole import with a generic 500 and no field path.

1e10 IDR ≈ US$600k, 1e10 VND ≈ US$400k — reachable in normal use for those currencies.

## Fix

- **Schema** — widened every money/quantity column from `Decimal(18, 8)` to `Decimal(28, 8)` (`Account.cashBalance`, `Holding.quantity`, `HoldingTransaction.quantity` / `unitPrice`, `CashTransaction.amount`, `RecurringCashTransaction.amount`, `RecurringInvestment.amount`, `PriceCache.price`, `StockWatchItem.recordPrice`, `Goal.targetAmount`) and the three `NetWorthSnapshot` aggregates from `Decimal(18, 2)` to `Decimal(28, 2)`.
- **Validators** — `CRUD_DECIMAL_ABS_MAX` raised to `1e15`. The DB now holds 20 integer digits, so it is no longer what binds: every amount crosses the wire as a JSON `number` and an IEEE double keeps ~15.95 significant digits, so 1e15 is the largest magnitude at which the fractional part is still meaningful. The same bound now applies to the import `decimalSchema` on **both** its string and number branch, so an oversized backup value is a 400 with a field path (`accounts.0.cashBalance`) instead of a Prisma 500.
- **Docs** — `docs/DATABASE.md` gains a short "Numeric precision" section recording the column widths and the wire-format ceiling.

## Migration notes

`prisma/migrations/20260901215445_widen_money_columns/migration.sql` — 12 `ALTER COLUMN … SET DATA TYPE` statements. Widening a `numeric` is a metadata-only change in PostgreSQL: no table rewrite, no data conversion, no downtime.

**Self-hosters must run `pnpm exec prisma migrate deploy`** (the Docker Compose `migrate` service and `build:vercel` already do this) before running a build that accepts the new range.

## Tests

`pnpm test:unit` (1090 passed) · `pnpm format:check` · `pnpm lint` · `pnpm typecheck` — all clean. `pnpm test:integration` ran locally against `asset_app_asset_tracker_test` after `migrate deploy`: 27 passed.

`tests/unit/validators.test.ts`:
- The old `rejects 1e10 for %s` table was **updated, not deleted** — it is now parametrized by value and asserts both boundaries (`1e15` rejected, `999_999_999_999_999` accepted) for `createAccountSchema.cashBalance`, `updateAccountSchema.cashBalance`, holding quantity/unitPrice, holding-transaction quantity, cash-transaction amount, both recurring amounts, goal targetAmount and stock-watch recordPrice.
- New `dataImportSchema` cases: a numeric and a decimal-string `cashBalance` just under the maximum are accepted; `1e15` and `"1000000000000000"` are rejected with message `Value exceeds supported precision` at path `["accounts", 0, "cashBalance"]`.
- New option-strike cases at its own boundary.

## Decisions

- **`ExchangeRate.rate` untouched** — a separate change widens its scale, as noted in the issue.
- **`Holding.strike` stays `Decimal(18, 4)`** (14 integer digits, max ~1e14). Raising the shared validator cap to 1e15 while leaving that column narrow would have reopened exactly this bug at a higher magnitude, so strike now carries its own `STRIKE_DECIMAL_ABS_MAX = 1e14` bound in both the CRUD and the import schema. Three lines, and it restores the invariant that everything the validator accepts fits its column.
- **1e15, not the column width (1e20)** — the API contract is JSON, and a double cannot represent 1e16 + 0.5. Accepting values the wire format silently rounds would trade a loud 500 for quiet corruption.
- The migration was generated with `prisma migrate dev` against a **scratch database**, because the shared local dev DB reported drift on an unrelated earlier migration; `prisma migrate diff --from-config-datasource --to-schema` against that fully migrated scratch DB reports "No difference detected".

Fixes #735

### DIFF
--- a/docs/DATABASE.md
+++ b/docs/DATABASE.md
@@ -9,6 +9,12 @@ astt uses PostgreSQL through Prisma. Standard PostgreSQL connections use `@prism
 
 For local development, `.env.example` contains URLs for the PostgreSQL service in `docker-compose.yml`.
 
+## Numeric precision
+
+Balances, amounts, quantities and prices are `DECIMAL(28, 8)`; net-worth snapshot aggregates are `DECIMAL(28, 2)`. Option strikes stay `DECIMAL(18, 4)` and exchange rates `DECIMAL(18, 8)`.
+
+The API accepts a smaller range than the columns hold: every amount crosses the wire as a JSON `number`, and an IEEE double keeps roughly 15.95 significant digits, so the validators in `src/lib/validators.ts` reject any magnitude at or above `1e15` (`1e14` for option strikes) with a 400. That headroom means an accumulating `cashBalance` cannot overflow its column.
+
 ## Creating migrations
 
 Create schema changes against a disposable development database:
--- a/prisma/migrations/20260901215445_widen_money_columns/migration.sql
+++ b/prisma/migrations/20260901215445_widen_money_columns/migration.sql
@@ -1,0 +1,32 @@
+-- AlterTable
+ALTER TABLE "Account" ALTER COLUMN "cashBalance" SET DATA TYPE DECIMAL(28,8);
+
+-- AlterTable
+ALTER TABLE "CashTransaction" ALTER COLUMN "amount" SET DATA TYPE DECIMAL(28,8);
+
+-- AlterTable
+ALTER TABLE "Goal" ALTER COLUMN "targetAmount" SET DATA TYPE DECIMAL(28,8);
+
+-- AlterTable
+ALTER TABLE "Holding" ALTER COLUMN "quantity" SET DATA TYPE DECIMAL(28,8);
+
+-- AlterTable
+ALTER TABLE "HoldingTransaction" ALTER COLUMN "quantity" SET DATA TYPE DECIMAL(28,8),
+ALTER COLUMN "unitPrice" SET DATA TYPE DECIMAL(28,8);
+
+-- AlterTable
+ALTER TABLE "NetWorthSnapshot" ALTER COLUMN "totalAssets" SET DATA TYPE DECIMAL(28,2),
+ALTER COLUMN "totalLiabilities" SET DATA TYPE DECIMAL(28,2),
+ALTER COLUMN "netWorth" SET DATA TYPE DECIMAL(28,2);
+
+-- AlterTable
+ALTER TABLE "PriceCache" ALTER COLUMN "price" SET DATA TYPE DECIMAL(28,8);
+
+-- AlterTable
+ALTER TABLE "RecurringCashTransaction" ALTER COLUMN "amount" SET DATA TYPE DECIMAL(28,8);
+
+-- AlterTable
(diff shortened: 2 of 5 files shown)


mike840609:
The schema/migration direction is sound and the validator/import coverage is good. Two wording/details should be tightened before merge:

1. `Decimal(28,8)` does not make an accumulating `cashBalance` literally impossible to overflow; repeated individually-valid increments could still exceed the column eventually. The docs should say the wider type makes overflow impractical under realistic usage, not impossible.
2. The `1e15` API bound is a reasonable magnitude guard for JSON numbers, but near that magnitude JavaScript `number` no longer preserves 8 fractional decimal places. Avoid wording that implies full `Decimal(28,8)` fractional precision survives the JSON wire format at values near the ceiling.

Code-wise I did not find a blocker. After those doc/comment clarifications, this looks merge-ready. Also note #742 touches `schema.prisma` and assumes the pre-#740 money-column scales in one test, so #742 should be rebased after this PR lands.

PR #741, opened 2026-09-01
fix(transactions): reverse a DCA buy's recorded cash debit exactly

### LINKED ISSUE #734: Deleting a DCA-generated BUY reverses the shares but not the cash debit
## Where

`src/lib/services/recurring-investment-service.ts:211-221` (materialization debits `cashBalance`) vs `src/app/api/accounts/[id]/transactions/[transactionId]/route.ts:256-293` (DELETE of a holding transaction only applies the quantity delta).

## Problem

Manual holding transactions never move cash, so DELETE not touching cash is consistent for them. DCA materialization is the exception: it debits `amount` from the account's cash for every BUY it posts. Deleting one of those rows (`recurringId` set) removes the shares but leaves the cash debit in place, so the account is permanently short by `amount` and the ledger no longer explains the balance.

The same asymmetry applies to editing the quantity of a DCA BUY via PATCH.

## Expected

When the deleted/edited holding transaction carries a `recurringId` (or `materializedAt`-style provenance), reverse the proportional cash debit inside the same `$transaction`, or reject the mutation with a message pointing the user at a manual cash adjustment.

## Severity

Medium — only reachable by deleting a generated row, but leaves balance and ledger inconsistent.

mike840609:
## Problem

Manual holding transactions never move cash, so `PATCH`/`DELETE` on
`/api/accounts/[id]/transactions/[transactionId]` only ever applied a quantity
delta. DCA materialization is the one exception: `recurring-investment-service.ts`
posts a BUY `HoldingTransaction` **and** debits the account's `cashBalance` by the
rule's `amount` in the same `$transaction`. Deleting or shrinking that BUY removed
the shares but left the debit in place, so the account stayed permanently short and
the ledger no longer explained the balance.

Fixes #734

## Why this was reworked

The first revision identified generated rows by `recurringId` and recomputed the
cash as `quantity x unitPrice x today's FX rate`. Review held it, correctly:

1. `HoldingTransaction.recurringId` is `onDelete: SetNull`. Once the recurring
   rule is deleted, a generated BUY is indistinguishable from a manual one and its
   debit is never reversed — the ledger-integrity bug was only half fixed.
2. A cross-currency reversal priced at today's rate can return a different amount
   than was actually spent.

Both are consequences of deriving, at reversal time, facts that were only known at
posting time. So the fix now **records those facts when the row is posted**.

## Design

`HoldingTransaction` gains two nullable columns:

- **`materializedAt DateTime? @db.Timestamptz(3)`** — durable generation
  provenance, exactly mirroring `CashTransaction`. Unlike `recurringId` it
  survives rule deletion, so a generated BUY stays recognizable forever.
- **`cashDebit Decimal? @db.Decimal(28, 8)`** — the exact cash the row removed
  from the account, **in the account's currency**. The materializer's debit is
  `rule.amount` per occurrence, so that is what each row stores.

No index: nothing queries either column (unlike `CashTransaction.materializedAt`,
which analysis floors by). No `materializedAtEstimated` twin either — this
timestamp is provenance only; nothing reads or displays it as an instant.

**Materializer** (`recurring-investment-service.ts`) sets `materializedAt` to the
run instant and `cashDebit` to `rule.amount` on every row `createMany` posts,
inside the existing reservation/CAS transaction.

**Reversal** (`transactions/[transactionId]/route.ts`):

- a row is generated when `materializedAt != null || recurringId != null`, so the
  reversal keeps working after the rule is gone — the case the review called out;
- **DELETE** with `cashDebit` present credits back that stored amount directly:
  no FX lookup, no recomputation, exact even when the holding's currency differs
  from the account's;
- **PATCH** resize scales the stored debit by the quantity ratio
  (`cashDebit x (newQty - oldQty) / oldQty`, in `Decimal`, snapped to 8 dp) and
  **writes the scaled debit back on the row** in the same `$transaction` — a
  resize that only moved cash would otherwise leave a later delete refunding the
  pre-resize amount and stranding the difference;
- **PATCH** still returns **400** on a type change (a BUY-shaped debit has no
  meaningful reversal as a SELL or EDIT), and both verbs still fail closed with
  **409** when no amount can be derived, writing nothing.

### The legacy fallback, and why it stays approximate

Rows generated before this migration have no `cashDebit`, so they keep the old
path: `quantity x unitPrice`, converted at today's rate. That is documented as the
legacy fallback in code, and it cannot be made exact retroactively:

- `unitPrice` is stored in the holding's currency and no rate snapshot was ever
  taken, so the occurrence-day rate is gone;
- the rule's `amount` is not a substitute — it may have been edited since, and the
  rule may not exist at all any more.

Same-currency rows (the common case) short-circuit to rate 1 and are exact.

## Migration

`20260902000000_add_holding_materialization_provenance`

```sql
ALTER TABLE "HoldingTransaction"
  ADD COLUMN "materializedAt" TIMESTAMPTZ(3),
  ADD COLUMN "cashDebit" DECIMAL(28,8);

UPDATE "HoldingTransaction"
SET "materializedAt" = "createdAt"
WHERE "recurringId" IS NOT NULL;
```

**Self-hosters:** additive and online — two nullable columns plus one backfill
`UPDATE`, no rewrite of existing values, no downtime step. The backfill is the same
compatibility move the `CashTransaction` provenance migration made: rows still
linked to a rule are known-generated, and the legacy materializer wrote
`createdAt` = the occurrence day, the best recoverable bound. Rows whose rule was
already deleted have `recurringId = NULL` and are indistinguishable from manual
rows, so they stay unclassified rather than guessed at. `cashDebit` is left NULL
for every pre-existing row on purpose (see above), so they keep the approximate
reversal.

Hand-written per repo convention, then verified by replaying the whole migration
directory into a throwaway database and diffing it against the schema
(`prisma migrate diff --from-config-datasource --to-schema` → *No difference
detected*).

## Backup round-trip

The export dumps whole models, so both fields flow out automatically; the export
`version` is bumped **1.4 → 1.5**. `dataImportSchema` accepts `materializedAt`
(nullable ISO datetime) and `cashDebit` (nullable bounded decimal, same 1e15
ceiling as every other imported amount), and the import route maps them.

An older backup still imports unchanged, and recovers what is knowable: a holding
transaction that carries a `recurringId` gets `materializedAt = createdAt` — the
same bound the migration backfills, so a user who exported before upgrading and
imported after ends up in the same state as one who simply migrated. `cashDebit`
stays null there.

## Tests

`tests/unit/account-ledger-routes.test.ts`

- DELETE with `cashDebit` reverses exactly that amount, cross-currency, and
  **asserts the rate loader is never called** (the empty rate map would fail
  closed if it were)
- DELETE still reverses when `recurringId` is null but `materializedAt` is set —
  the review's case
- PATCH resize scales the stored debit proportionally, both signs, no FX
- PATCH resize writes the scaled `cashDebit` back, and the resized row then
  deletes for the full resized amount
- legacy row (`cashDebit` null, `recurringId` set): same-currency, cross-currency
  at today's rate (asserting the loader *is* used), 409 on unresolvable rate, 409
  on missing `unitPrice`, proportional resize, and no `cashDebit` write-back
- 409 when a stored debit cannot be scaled by a zero quantity
- manual row moves no cash on delete or resize
- 400 on a generated buy's type change

`tests/unit/recurring-investment-service.test.ts` — materialization stores
`materializedAt` = the run instant and `cashDebit` = `rule.amount`; per-row (not
scaled) on catch-up and on an idempotent partial insert; in the account's currency
for a cross-currency rule.

`tests/unit/calendar-backup-route.test.ts` — v1.5 round-trip preserves both fields
through the id remap; an older backup recovers `materializedAt` from `createdAt`
for a generated row and leaves a manual row's provenance null; export version
asserted as 1.5.

`tests/unit/validators.test.ts` — both fields survive parsing, absent fields stay
`undefined`, and an oversized `cashDebit` is rejected by the schema rather than at
write time.

`pnpm test:unit` (1118 passed), `format:check`, `lint`, `typecheck` clean;
`pnpm test:integration` (27 passed) against a local `_asset_tracker_test` database
with the migration applied via `prisma migrate deploy`.

## Decisions

- **Two columns, not one.** Provenance alone would keep the reversal firing after
  a rule is deleted but leave it approximate; the amount alone would still be
  invisible once `recurringId` is nulled. The review asked for both, and both are
  needed.
- **`cashDebit` in the account's currency.** That is the currency `cashBalance` is
  denominated in and the currency the debit was applied in, so the reversal never
  needs an exchange rate at any point in the row's life.
- **Resize rewrites `cashDebit`.** Necessary for the invariant "the stored debit is
  what this row currently owes the account"; the delta is derived from the old and
  new stored amounts so balance and row round identically.
- **No `materializedAtEstimated`.** `CashTransaction` needs it because analysis
  surfaces that timestamp; here it is pure provenance, never read as an instant.
- **No contract multiplier.** DCA rules only ever target plain symbols, never
  OPTION holdings.

### DIFF
--- a/prisma/migrations/20260902000000_add_holding_materialization_provenance/migration.sql
+++ b/prisma/migrations/20260902000000_add_holding_materialization_provenance/migration.sql
@@ -1,0 +1,26 @@
+-- Keep DCA-generated buy provenance after its source rule is deleted, and
+-- record the exact cash each materialized buy removed from the account.
+--
+-- HoldingTransaction.recurringId is ON DELETE SET NULL, so deleting a recurring
+-- investment made its generated buys indistinguishable from manual ones and
+-- their cash debit was never reversed. "materializedAt" survives that deletion.
+-- "cashDebit" is stored in the ACCOUNT's currency (the one the cash balance is
+-- denominated in), so a reversal replays the original amount instead of
+-- recomputing quantity x unitPrice at today's exchange rate.
+ALTER TABLE "HoldingTransaction"
+  ADD COLUMN "materializedAt" TIMESTAMPTZ(3),
+  ADD COLUMN "cashDebit" DECIMAL(28,8);
+
+-- Same compatibility move the CashTransaction provenance migration made: rows
+-- still linked to a rule are known-generated, and the legacy materializer wrote
+-- createdAt = the occurrence day, the best recoverable bound. Rows whose rule
+-- was already deleted have recurringId = NULL and are indistinguishable from
+-- manual rows, so they stay unclassified rather than guessed at.
+--
+-- "cashDebit" is deliberately left NULL for every legacy row: the amount spent
+-- per occurrence is not recoverable, because the rule's "amount" may have been
+-- edited since the row was posted. Those rows keep the approximate
+-- quantity x unitPrice reversal.
+UPDATE "HoldingTransaction"
+SET "materializedAt" = "createdAt"
+WHERE "recurringId" IS NOT NULL;
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -115,6 +115,17 @@ model HoldingTransaction {
   recurringId    String?
   recurring      RecurringInvestment? @relation(fields: [recurringId], references: [id], onDelete: SetNull)
   occurrenceDate DateTime?            @db.Date
+  // Durable DCA-generation provenance, mirroring CashTransaction. Unlike
+  // recurringId (SetNull) this survives rule deletion, so a generated BUY stays
+  // recognizable — and its cash reversible — after the rule is gone.
+  // `cashDebit` is the exact cash the materialization removed from the account,
+  // in the ACCOUNT's currency, so a reversal never re-derives it from price and
+  // today's FX. NULL on manual rows and on rows generated before these columns
+  // existed (their per-occurrence amount is unrecoverable: the rule's `amount`
+  // may have been edited since). No `materializedAtEstimated` twin here: this
+  // timestamp is provenance only — nothing reads or displays it as an instant.
+  materializedAt DateTime?            @db.Timestamptz(3)
+  cashDebit      Decimal?             @db.Decimal(28, 8)
 
   @@unique([recurringId, occurrenceDate])
   @@index([holdingId, createdAt(sort: Desc)])
--- a/src/app/api/accounts/[id]/transactions/[transactionId]/route.ts
+++ b/src/app/api/accounts/[id]/transactions/[transactionId]/route.ts
@@ -1,4 +1,5 @@
 import { revalidateTag } from "next/cache";
+import type { TransactionType } from "@/generated/prisma/client";
 import { Decimal } from "@/generated/prisma/internal/prismaNamespace";
 import { prisma } from "@/lib/prisma";
 import { updateTransactionSchema, updateCashTransactionSchema } from "@/lib/validators";
@@ -10,6 +11,7 @@ import {
   normalizeHoldingTransactionQuantity,
   toDbMoneyDelta,
 } from "@/lib/services/balance";
+import { getFreshExchangeRates, resolveRate } from "@/lib/services/exchange-rate-service";
 import { ok, failure, validationError } from "@/lib/api-responses";
 import { withAuth } from "@/lib/api-handler";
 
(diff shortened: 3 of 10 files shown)


mike840609:
The current implementation fixes the common case, but I would hold this PR until the generated-buy provenance is durable enough to support an exact reversal.

The main correctness gap is `HoldingTransaction.recurringId @relation(... onDelete: SetNull)`: once the recurring rule is deleted, a generated BUY becomes indistinguishable from a manual BUY, so deleting/resizing that transaction later no longer reverses the cash debit. The PR documents this limitation, but it means the underlying ledger-integrity issue is only partially solved.

There is a second approximation for cross-currency DCA: reversal uses today's FX rate because the transaction does not retain the original account-currency cash debit / FX snapshot. That can return a different amount of account cash than was originally spent.

I recommend storing durable generation provenance plus the exact original cash debit (or equivalent FX/cash snapshot) on `HoldingTransaction`, then reversing that stored amount directly. The existing tests and logic are otherwise solid, and CI/E2E are green.

PR #742, opened 2026-09-01
fix: data-correctness audit low-priority follow-ups

### LINKED ISSUE #736: Data-correctness audit: low-priority follow-ups
Low-priority findings from the 2026-09 data-correctness audit, grouped so they can be picked off individually. None of them corrupt stored data.

- [ ] **History and Analysis disagree on deleted accounts.** `normalizeBreakdown` (`src/lib/services/history-service.ts:165`) bails to the stored aggregates when a breakdown entry references an account that no longer exists, so the History trend still includes that account's value. `getRawHistoryWithBreakdown` (`:478-483`) converts every entry, but `aggregateCategoryHistory` (`src/lib/services/analysis-service.ts:326-329`) iterates the live `accounts` list, so the Analysis category chart drops it. The two pages show different totals for the same snapshot.
- [ ] **FX freshness gate is satisfied by inverse rows written for another base.** `getPersistedFreshRefreshAt` (`src/lib/services/exchange-rate-service.ts:126-134`) keys on `fromCurrency`, and every refresh also writes inverse rows (`:392-400`). After a USD refresh, `refreshExchangeRates("TWD")` sees the fresh `TWD_USD` row and skips its own fetch. The USD cross covers most pairs so this is mostly a wasted-fetch/precision nit, but the "skipped fresh" toast can be misleading.
- [ ] **Inverse rates lose precision at 8 decimals for very weak currencies.** `1 / rate` for LBP (~89,500/USD) or IRR is stored as `0.00001117`, ~0.3% off. `resolveRate` prefers the direct row, so a user whose base is one of these currencies converts with the truncated value. Storing the inverse at higher scale, or deriving it on read from the forward rate, removes the error.

mike840609:
The three low-priority findings from the 2026-09 data-correctness audit. None of them corrupted stored data; two were silent wrong numbers and one was a wasted fetch with a misleading toast.

Rebased onto `master` after #740 (money columns widened to `Decimal(28, 8)`, import bound raised to `1e15`, "Numeric precision" section added to `docs/DATABASE.md`). The review fixes that rebase required are in the last section.

Fixes #736

---

## 1. History and Analysis disagreed on deleted accounts

**Problem.** When a snapshot's breakdown references an account that no longer exists, `normalizeBreakdown` (`src/lib/services/history-service.ts`) cannot classify it — guessing `ASSET` vs `LIABILITY` could invert net worth — so it bails to the stored aggregates and the History trend keeps the value. `aggregateCategoryHistory` and `computePerformanceAttribution` instead iterate the live `accounts` list, so the Analysis composition charts drop it. Both behaviours are individually correct; the two pages just reported different totals for the same day with nothing explaining why.

**Fix.** Surface the disagreement instead of trying to reconcile it.

- `getRawHistoryWithBreakdown` now also returns `unattributedDates: string[]` — the `YYYY-MM-DD` of every deduped snapshot carrying at least one breakdown entry with no live account. `SnapshotBreakdown` and the History trend are unchanged.
- `computeAnalysisRangeSeries` folds that into a per-range `hasUnattributedAccounts` boolean (any affected date `>= rangeStartIso`), added to `AnalysisRangeSeries` in `src/lib/analysis-contract.ts`.
- The Analysis **Composition** section renders one muted note when the flag is set, in both locales.

**Files.** `src/lib/services/history-service.ts`, `src/lib/services/analysis-series-service.ts`, `src/lib/analysis-contract.ts`, `src/components/analysis/analysis-view.tsx`, `messages/en-US.json`, `messages/zh-TW.json`.

**Tests.** `tests/unit/history-service.test.ts` (4 cases: none / one affected date / several deleted accounts deduped and sorted / empty and malformed breakdowns), `tests/unit/analysis-series-service.test.ts` (3 cases: flag false when clean, scoped correctly across all five ranges, inclusive on the range-start boundary), `tests/unit/analysis-unattributed-note.test.ts` (new — the note is gated on the flag, uses the muted pattern, and exists in both bundles).

**Decisions.**
- The note sits under the **Composition** section heading rather than beside the category chart alone, because the wording covers both the category *and* attribution charts, which live in that same section.
- Styling reuses the existing muted subtitle pattern (`text-xs text-muted-foreground`, plus `mt-1 max-w-prose`) — no new colours or tokens.
- The component test is source-text based, matching how `issue-592-components.test.ts` and `analysis-view-deferred-observer.test.ts` already assert on components (the unit suite runs in the `node` environment; there is no jsdom or testing-library in the project).

---

## 2. FX freshness gate satisfied by inverse rows written for another base

**Problem.** `refreshExchangeRates` writes the fetched forward rows (`base_X`) and derived inverse rows (`X_base = 1/rate`) in one upsert that stamped **every** row with a SQL-side `NOW()`. `getPersistedFreshRefreshAt` gates the external fetch on `MAX("updatedAt") WHERE "fromCurrency" = base`, so a row written purely as a by-product of another base's refresh answered for a currency that had never been fetched: after a USD refresh, `refreshExchangeRates("TWD")` saw the fresh derived `TWD_USD` row, logged `rates.refresh.skipped_fresh_db`, and served a `1/rate` value instead of the upstream TWD cross.

**Fix.** The stamp is now a per-row parameter — the refresh instant for fetched rows, the epoch for derived ones — and the conflict clause is `"updatedAt" = GREATEST("ExchangeRate"."updatedAt", EXCLUDED."updatedAt")`, so a derived write landing after a genuine one cannot push a real refresh timestamp backwards.

**Files.** `src/lib/services/exchange-rate-service.ts`.

**Tests.** `tests/unit/exchange-rate-service.test.ts` — the SQL carries `GREATEST` and a `::timestamptz` per row, the fetched row carries the refresh instant and the derived row the epoch; a derived row does not make `refreshExchangeRates(otherBase)` skip as fresh (the mocked `exchangeRate.aggregate` now really computes `MAX(updatedAt)` per `fromCurrency`, and the test replays exactly what the service wrote); a base whose own forward rows are recent still skips.

**Decisions.**
- The timestamp is passed as an ISO parameter rather than left as a SQL-side `NOW()`, because `NOW()` cannot be varied per row. Side effect: the stamp is now app-clock rather than DB-clock, which is actually more consistent with the readers — `getPersistedFreshRefreshAt` and the in-process `lastRateRefreshAt` memo both compare it against `Date.now()`.
- I checked every other reader of `ExchangeRate.updatedAt`: the Settings rates freshness badge takes `MAX` over all rows (`findFirst` ordered by `updatedAt` desc), and `/api/exchange-rates`, the Settings data export, `/api/health` and the three warm-on-write paths never select the column. Nothing uses `MIN` or a per-row check, so the epoch rows stay invisible to them and no adjustment was needed.
- `rate = EXCLUDED.rate` stays unconditional — the existing last-write-wins tradeoff for the *value* is unchanged; only the freshness stamp's meaning is tightened.

---

## 3. Inverse rates lost precision at 8 decimals for very weak currencies

**Problem.** `resolveRate` prefers a direct row over re-deriving one, and `ExchangeRate.rate` was `@db.Decimal(18, 8)`. The persisted inverse of a very weak currency lost most of its significant digits: `1/89,500` (LBP per USD) stored as `0.00001117`, ~0.3% off, so a user based in LBP or IRR converted every value with the truncated number.

**Fix.** Widen **only** `ExchangeRate.rate` to `@db.Decimal(28, 16)`, and compare rate changes at 12 decimals instead of 8 (`decimalChangedAtDbScale` was pinned to the old column scale and would otherwise report a genuinely-changed inverse as unchanged). 12 is past the 8th decimal but short of the double's own round-trip noise at the new scale.

`Decimal(28, 16)` is deliberately wider than the `Decimal(28, 8)` the rest of the money columns use after #740: exchange rates are the one place where an inverse is persisted, so the *scale* — not the precision — is what binds.

**Files.** `prisma/schema.prisma`, `prisma/migrations/20260902000000_widen_exchange_rate_scale/migration.sql`, `src/lib/services/exchange-rate-service.ts`, `docs/DATABASE.md`.

**Tests.** `tests/unit/exchange-rate-precision.test.ts` (new — the schema declares `Decimal(28, 16)`, and the migration widens that one column while touching no other table) and `tests/unit/exchange-rate-service.test.ts` (the derived inverse reaches the DB with more than 8 decimals; a rate differing only past the 8th decimal counts as changed; an identical rate still counts as unchanged).

**Decisions.**
- The migration is hand-written. `prisma migrate dev` refused to run: the shared local development database has pre-existing checksum drift on `20260820001418_add_calendar_earnings_watch` and Prisma wanted a full reset, which would have destroyed another worktree's data. Instead the whole migration directory was replayed into a throwaway database — `prisma migrate diff --from-config-datasource --to-schema` then reports **"No difference detected"**, and `information_schema` confirms the column lands as `numeric(28, 16)` with `Account.cashBalance` still `numeric(28, 8)`. The scratch database was dropped afterwards.
- Only `ExchangeRate.rate` is touched; every other money column keeps the `Decimal(28, 8)` / `Decimal(28, 2)` that #740 gave it.
- `decimalChangedAtDbScale` is tested through `refreshExchangeRates`'s `changed` count rather than by exporting the helper, so no API surface is added just for a test.
- `String(rate)` for the `::numeric` parameter is unchanged — it already emits full double precision, and there is now a test locking that.

### Migration note for self-hosters

This PR adds one migration, `20260902000000_widen_exchange_rate_scale`: `ALTER TABLE "ExchangeRate" ALTER COLUMN "rate" TYPE DECIMAL(28, 16)`.

**This one is not a metadata-only change.** The scale goes from 8 to 16, not just the precision, so — unlike #740's same-scale widening — PostgreSQL rewrites the table and holds `ACCESS EXCLUSIVE` on it for the duration. `ExchangeRate` stores one row per currency pair (tens of rows on a normal install), so the lock is brief in practice, but it does block reads and writes of that table while it runs, and it needs transient disk for the rewritten copy.

No data is lost, no other table is touched, and no application change is required beyond the usual `prisma migrate deploy` on startup. Existing rows keep their stored (truncated) values; they pick up full precision on the next FX refresh.

---

## Review fixes

Applied after review, on top of the rebase onto #740.

1. **`tests/unit/exchange-rate-precision.test.ts` no longer asserts other models' scales.** It asserted `Account.cashBalance` was `Decimal(18, 8)`, which #740 made false. The assertion's real intent was "this migration touches only `ExchangeRate`", which the migration-content assertions (`not.toContain('ALTER TABLE "Account"')` and friends) already prove — so the assertion is deleted rather than renumbered, and the test stops coupling itself to unrelated columns.
2. **`docs/DATABASE.md` "Numeric precision" updated.** #740's new section said exchange rates are `DECIMAL(18, 8)`; this PR makes them `DECIMAL(28, 16)`, with a clause on why they are wider than the money columns.
3. **Migration note corrected** in both the PR body (above) and the migration SQL comment: `Decimal(18,8) -> Decimal(28,16)` changes the scale, so it is a table rewrite under `ACCESS EXCLUSIVE`, not the metadata-only widening #740 correctly claimed for its own same-scale change.

Items 4 and 5 close out the post-merge review of #740. Both are wording only — no behaviour change, and they are in a separate commit for that reason.

4. **"cannot overflow its column" softened.** Repeated individually-valid increments could still reach the ceiling eventually, so the widened column is described as a backstop that makes overflow impractical under realistic use, not as an impossibility proof.
5. **The `1e15` bound is presented as a magnitude guard, not a precision guarantee**, in that section and in the `CRUD_DECIMAL_ABS_MAX` comment in `src/lib/validators.ts`. A JSON `number` is an IEEE double, so `Decimal(28, 8)`'s eight fractional decimals do not survive the wire anywhere near `1e15`. Both figures were verified numerically before being written down:
   - All eight fractional decimals are distinguishable only below `2^26` (`67,108,864`, ~`6.7e7`), where one ulp is `2^-27` ≈ `7.45e-9`. At and above `2^26` the ulp is ≈ `1.49e-8`, coarser than `1e-8`.
   - At `1e15` the ulp is exactly `0.125` (`1e15 + 0.0625 === 1e15`, `1e15 + 0.125 !== 1e15`) — whole eighths, no decimals.
   - `1e15` still sits an order of magnitude below `2^53` (`9,007,199,254,740,992`), where doubles stop representing even whole units exactly, which is what the bound actually buys.

---

## Validation

`pnpm test:unit` (121 files, 1119 tests), `pnpm format:check`, `pnpm lint`, `pnpm typecheck` — all clean. `pnpm test:integration` also run against a local `_asset_tracker_test` database with every migration applied (2 files, 27 tests passing). The migration directory was separately replayed into a throwaway database, where `prisma migrate diff --from-config-datasource --to-schema prisma/schema.prisma` reported no difference.


mike840609:
Reviewed all three follow-ups. The deleted-account disclosure and the FX freshness-stamp fix are both well targeted. The inverse-rate precision change is also directionally correct, but this PR should be rebased after #740 before merge:

- `tests/unit/exchange-rate-precision.test.ts` currently asserts that `Account.cashBalance` remains `Decimal(18, 8)`, which will be false after #740 widens money columns.
- `docs/DATABASE.md` from #740 will say exchange rates are `Decimal(18,8)` unless updated after this PR changes them to `Decimal(28,16)`.
- The migration note should avoid calling `Decimal(18,8) -> Decimal(28,16)` a pure metadata-only widening with no row rewrite: the scale changes from 8 to 16, not just precision.

The application logic itself looks good, and CI/E2E/Docker smoke are green. Recommended sequence: merge #740, rebase this branch, fix the combined-state test/docs/migration wording, rerun CI, then merge.

